PR #27, opened 2017-06-05
Initial redirect implementation of redirection, connect() and getuid()

riataman:
This is an initial implementation of the redirect funcionality.

I created two pieces of infrastructure, one is the

´int get_tracing_enabled();´
´int set_tracing_enabled(int enabled);´

Functions, which disable tracing. This is useful to avoid introducing noise to the trace with our own operations. Also to avoid loops when our own functions calls themselves call the functions we are tracing. Think of the getuid redirector which goes getuid->get_redirect->fopen->getuid and then we are looping.

The other function is the

 get_redirect (const char *function, ...)´

function, which takes the name of a function and searches for it in the config file. and then returns the parameters of the redirection.

Its an initial implementation to get started and still needs some improvements. It'll try to open /etc/retrace.conf (there's an example in the commit retrace.conf.example) (it would be nice to also search ~/.retrace.conf.

It also needs some caching so it won't scan the file every single time. Also needs a way to check for several different redirects for the same function (think different ip/port address combinations to connect).

Anyway, this is an initial implementation and I think a good starting point.


ronaldtse:
@efiop @riataman  I've created #29 to deal with code formatting. An automated solution to me seems easier to enforce 👍

ronaldtse:
Since I don't have access to @riataman 's retrace/master, I've created #32 to apply clang-formatting.

ronaldtse:
#32 is merged and closed (hence #27 is merged).

PR #34, opened 2017-06-06
Adds cpp macros to reduce boilerplate

drystone:
I've added some macros to common.h

`RETRACE_DECL()` declares the redirection pointer and the getter prototype
`RETRACE_GETTER()` allocates the redirection pointer and defines the getter
`RETRACE()` adds a default handler
`RETRACE_FN()` adds a default handler but overrides the trace function

I'm sure it could be improved upon, but does reduce a lot of code repetition
Comments welcome!


ronaldtse:
Travis is now displaying this error:
```
cmockatest.c:43:1: error: ‘for’ loop initial declarations are only allowed in C99 mode
 for (char *p = "abcDEF"; *p; ++p)
```

On the other hand, are we going to use the prefix `rtr_` or `trace_`?

erikbor:
@drystone, I personally am not really keen on using macros. But I would like to have the opinion of the other developers whether we should follow this approach or not.

@drystone I do like the optimisations you made in fseek(), that's exactly the type of code review we want.

riataman:
I think something like this is going to be necessary anyway for the mac os port as their aproach in there is different. Will look how this integrates to that tonight (in about 8 hours :).

efiop:
These macros are useful, but I'm afraid at this stage of development they will become more of a hassle, as code base isn't anywhere near being stable, so we might have to work _around_ these macros instead of working _with_ them. So I have to agree with @erikbor on this one.

That being said, after we'll make the retrace do some of the designed things and code base will become a bit more stable, we will be able to go back and do a RFC to cleanup such common pieces of code. For now, these macros help with simple function wrappers where we just print stuff(and it does save us a LOT of common code), but it is about to change soon with adding of redirection functionality and I would prefer to take a closer look at macros at least after that is being done.

drystone:
@ronaldtse sorry about that ... I'll amend. I notice Travis is using GCC 4 and I'm GCC 6 - although I still don't get the warnings with --std={gnu,c}99!

@erikbor I agree, macros are a pain but without any form of overloading I think it's at least a typesafe solution

@riataman I'm interested to hear how you get on

@efiop and @everyone I agree the code is in flux and perhaps too soon to refactor so brutally. However, looking at issues such as tracing return values (which doesn't appear to be an issue any more) and configurable redirection, I think a small amount of work will introduce a lot more functionality. Already the macros don't fit all cases - the original functions are still working fine - but for the simple cases such as tolower it still makes sense to me. I'm easy however - it was a fun exercise.

I think the RETRACE_DECL and RETRACE_GETTER are useful though, as they introduce a mechanism with which to get the function pointers (get_real_*()) once per program invocation - it avoids the repeated calls to dlsym() at the top of every function (and for every char traced via trace_printf_str - although I'm tempted to look at optimizing this)

ronaldtse:
Thanks @drystone , I'm sure this work will come back another way and won't be wasted 👍

Issue #33, opened 2017-06-05
turn spaces indentation to tabs in .clang-format


erikbor:

ronaldtse:
@erikbor we've tried using tabs in riboseinc/rnp but it's been too difficult to coordinate. GitHub also doesn't display tabs consistently. Only spaces we've found to be consistent across platforms/editors.

erikbor:
Spaces are not used for indentation in *BSD nor Linux (Linux kernel coding style). I don't think that GitHub's inability to properly display tabs consistently should affect our usage.
I understand the tab/spaces discussion cannot be easily resolved, but maybe we can poll the use with the people who contribute to retrace to see what preference they have and use that to decide what type of indentation we use?

efiop:
I agree with @erikbor. Even though github sometimes screw ups tabs in pull-requests, merged code stills looks good. Using tabs is common for majority of C projects and it is part of both Linux Kernel Coding Style and BSD Coding Style. A bunch of hooks, travis and verbose output about code style should be plenty to enforce tabs without hiccups. We can also add a note about tab/indentation size(8 character width) to our Developer's Guide.

ronaldtse:
Alright, I was tempted to say that not everyone here is a kernel hacker, but I might be wrong here 👍 

Feel free to update it.

erikbor:
As discussed, we change space indentation to tab indentation. Solved by aee6ca6b6cba201d3420006b5d205371b341e5a1

PR #37, opened 2017-06-06
set clang-format IndentWidth to 8 and changed space indents to tabs

erikbor:


erikbor:
This solves "turn spaces indentation to tabs in .clang-format #33"

PR #42, opened 2017-06-07
Initial macOS port

riataman:
This will build a dylib lib in macOS.

Is has to be used like this: 

DYLD_INSERT_LIBRARIES=./retrace.dylib ./test-executable

I had to put some decoration macros, as in macOS we have to explicily declare the function that we want to override, with the function we want to override it with.

The macros in linux are basically boil down to noops.

I'll update the documetation tomorrow as its getting a bit late today.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,22 @@
+OS 		= $(shell uname)
 LD		= ld
 GCC		= gcc
 RM		= rm -f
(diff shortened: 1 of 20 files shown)


riataman:
Forgot to mention that I tested all of this in macOS sierra with the default compiler (clang).

erikbor:
Nice one Pablo! Thx

Issue #12, opened 2017-06-01
Add detailed info to connect(), e.g.: display the character arrays and structs in C-style


erikbor:

jjr840430:
@erikbor Can you please review https://github.com/riboseinc/retrace/blob/issue_20/sock.c?

jjr840430:
@erikbor 
I have implemented redirection of recv/send functions.
Please review this branch.

https://github.com/riboseinc/retrace/tree/issue_20

ronaldtse:
Shipped: v2 logs dereferenced buffer/struct content (`log_params` one-level derefs, `capture_buffer` for payloads) and `decode_http`/`decode_dns` produce structured protocol fields (cookbook 22).

Issue #10, opened 2017-06-01
Add exec* family


erikbor:

maheitsec:
Hello

I'd like to confirm exec* family functions.
execl
execv
execle
execve
execlp
execvp
Is that all?

erikbor:
execve() is already implemented so you can skip that one. Please implement the others and use a similar approach of displaying argv[] (where applicable) as we have done in execve().

cheers!

maheitsec:
int execl(const char *path, const char *arg0, ... /*, (char *)0 */);
I can get all parameters using va_list, but I can the real function using va_list as parameters.
So I'm going to make the char *argv[] and call execv function.
int execv(const char *path, char *const argv[]);
What do you think?

erikbor:
Yes good approach.
execl and execv both replace the calling process image with a new process image.

Thanks!

maheitsec:
Hello, Following functions were implemented.
execl, execv, execle, execve, execlp, execvp
I'm going to test these before merging to master.
Could you please let me know how to test these and test case exactly?

erikbor:
We use Cmocka for testing purposes: [Cmocka](https://cmocka.org/)
Please have a look at the existing test:
```sh
test/cmockatest.c
```
You can integrate your tests in there. For testing exec* I think we can use '/usr/bin/id' because that would also be a good test case for the getuid() redirection.

maheitsec:
Hello
I implemented 6 functions(execl, execv, execle, execve, execlp, execvp), but now I'm seeing new functions:
execvpe, fexecve, execveat
Should I implement them, too?

erikbor:
Implemented. Thanks!

Issue #9, opened 2017-06-01
Make Coverity Scan work


ronaldtse:
As title.

Link: http://scan.coverity.com/projects/riboseinc-retrace

efiop:
Hi @ronaldtse ! I see that retrace is registered in coverity, but I can't access it. You probably should make it public in settings or something.

wizzard:
Please add me as Maintainer to Coverity retrace project, I'll setup the project (please use my email)

ronaldtse:
Done @wizzard !

ronaldtse:
Sorry for the delay @efiop , missed your message 😉

ronaldtse:
Shipped: `.github/workflows/coverity.yml` runs the daily Coverity scan and uploads to scan.coverity.com.

Issue #8, opened 2017-06-01
Add accept()


erikbor:

maheitsec:
Added accept function in sock.c and sock.h

erikbor:
Great, thanks!

PR #48, opened 2017-06-07
Adds rtr_dlsym() to cache function ptrs

drystone:
This addresses issue #47 which should improve performance as well as avoiding infinite dlsym() loop

### DIFF
--- a/char.c
+++ b/char.c
@@ -32,8 +32,8 @@
 int
 RETRACE_IMPLEMENTATION(putc)(int c, FILE *stream)
 {
-	real_putc = dlsym(RTLD_NEXT, "putc");
-	real_fileno = dlsym(RTLD_NEXT, "fileno");
+	real_putc = rtr_dlsym(rtr_putc);
+	real_fileno = rtr_dlsym(rtr_fileno);
 	int fd = real_fileno(stream);
 
 	trace_printf(1, "putc(");
@@ -58,7 +58,7 @@ RETRACE_REPLACE (putc)
 int
 RETRACE_IMPLEMENTATION(toupper)(int c)
 {
-	real_toupper = dlsym(RTLD_NEXT, "toupper");
+	real_toupper = rtr_dlsym(rtr_toupper);
 	trace_printf(1, "toupper(\"%s\");\n", &c);
 	return real_toupper(c);
 }
@@ -68,7 +68,7 @@ RETRACE_REPLACE (toupper)
 int
 RETRACE_IMPLEMENTATION(tolower)(int c)
 {
-	real_tolower = dlsym(RTLD_NEXT, "tolower");
+	real_tolower = rtr_dlsym(rtr_tolower);
 	trace_printf(1, "tolower(\"%c\");\n", &c);
 	return real_tolower(c);
 }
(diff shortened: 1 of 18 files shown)


erikbor:
John, can you resolve the merge conflicts against the current master please? Thanks!

drystone:
@erikbor - do you think this is worthwhile? The lookup table is going to be much faster than dlsysm but it does mean a new member has to be added to the enum in common.h and a new lookup for each function hook in common.c ?

Also, I don't understand why the real_{func} variables are all global - although it does no harm as they will always contain the same value, it's unusual practice - especially as they are defined in header files.
I'd rather see local vars eg.

`rtr_fileno_t real_fileno = RETRACE_GET_REAL(fileno);`

erikbor:
@efiop @riataman can you please share your thoughts on this? thanks!

riataman:
As much as it would be nice to avoid a centralized registry of functions, I'm running into a deadlock that this would probably fix. I'm trying to launch firefox under retrace and I'm running into this.

For some reason the call from strcpy->dlsym gets into an error and that causes it to call into our tapped getenv which calls dlsym too, and since we are already in the middle of a call to dlsym this dies.

I can't duplicate this  by calling strcpy myself, so there's something else fishy going on, but I think this change (once updated again ...) would probably fix this.

#0  __lll_lock_wait () at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007ffff79be4af in _L_lock_1081 () from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x00007ffff79be428 in __GI___pthread_mutex_lock (mutex=0x555555771750)
    at ../nptl/pthread_mutex_lock.c:134
#3  0x0000555555567cc9 in ?? ()
#4  0x0000555555568e2d in calloc ()
#5  0x00007ffff77b1640 in _dlerror_run (operate=operate@entry=0x7ffff77b1130 <dlsym_doit>,
    args=args@entry=0x7fffffffbf60) at dlerror.c:141
#6  0x00007ffff77b1198 in __dlsym (handle=<optimized out>, name=<optimized out>) at dlsym.c:70
#7  0x00007ffff7bd40b8 in getenv (envname=0x55555556bdaa "MALLOC_OPTIONS") at env.c:49
#8  0x0000555555567d76 in ?? ()
#9  0x0000555555568e2d in calloc ()
#10 0x00007ffff77b1640 in _dlerror_run (operate=operate@entry=0x7ffff77b1130 <dlsym_doit>,
    args=args@entry=0x7fffffffd080) at dlerror.c:141
#11 0x00007ffff77b1198 in __dlsym (handle=<optimized out>, name=<optimized out>) at dlsym.c:70
#12 0x00007ffff7bd86f6 in strcpy (s1=0x7fffffffd11d "firefox-esr", s2=0x55555556acbb "libxul.so")
    at str.c:155
#13 0x0000555555558ca3 in _start ()

erikbor:
Ok John, please go ahead with this.

Thanks

drystone:
I've avoided a centralised function by extending the existing macros to wrap a statically stored function pointer.

The global real_{func} pointers are no longer available so we need to use locals.

It is possible two threads could simultaneously set the cached pointer but this should be harmless as I've used the __atomic family to avoid the possibility of reading half pointers.

PR #52, opened 2017-06-08
Exec family

maheitsec:
Added exec* family functions
execl
execv
execle
execve
execlp
execvp
execvpe
fexecve
execveat


erikbor:
Hi Ma He,
Can you please introduce some #if macOS statements for execvpe() and execveat() because these do not exist on macOS and building on macOS fails now.

Once the build also works on macOS then I can merge.

Thanks!!!

PR #57, opened 2017-06-09
Printf

drystone:


drystone:
Please wait on this as the Travis output differs from mine!

erikbor:
Sure, let me know when it's ready. Thanks.

drystone:
This is now working - the problem is that the trace_printf_str is broken. I'll fix this with the trace_printf pull request

PR #80, opened 2017-06-10
Build and loop fixes

riataman:
This brings everything back into shape. Builds and works in macOS and LInux.

There's a deadlock in linux when the traced program calls getpwuid that I haven't fixed, need to think about that one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,7 @@ install:
 
 script:
   - ./ci/main.sh
+
+os:
+  - linux
+  - osx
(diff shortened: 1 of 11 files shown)


riataman:
Also added macOS support to travis and a simple script that runs the tests I have.

I disabled the cmocka tests as the ones that tests where or not the function exists in the shared library don't make sense in macOS and the other ones are using fmemopen which is not available in macOS.

erikbor:
Thanks Pablo!

PR #84, opened 2017-06-11
Fixes

riataman:
This fixes all tracing stopping when multiple threads where modifying g_enable_tracing.

The fix is just to set g_enable_tracing as a per thread variable, as we are only protecting about our own code getting in loops. 

Also builds the tests using -fno-builtin so that we can actually tests our implementation.


riataman:
This is breaking macOS, checking it now

PR #93, opened 2017-06-13
Multithreading fixes

riataman:
This fixes threading stopping by itself on multithreaded apps by keeping track of the tracing enabled/disabled per thread. 

Includes a few misc fixes.


common.c
```diff
@@ -177,18 +196,52 @@ trace_dump_data(const void *buf, size_t nbytes)
...
+	pthread_key_create(&g_enable_tracing_key, NULL);
+}
+
+static int
+is_main_thread(void)
+{
+#ifdef __APPLE__
+	return pthread_main_np();
+#else
+	rtr_getpid_t real_getpid = RETRACE_GET_REAL(getpid);
```
efiop: No mixed declarations and code please.

common.c
```diff
@@ -52,8 +57,22 @@
...
+ * get us into a loop.
+ *
+ * g_enable_tracing_key logic is backwards, 0 (NULL)
+ * means tracing is enabled. 1 means tracing is disabled.
+ *
  **************************************************/
-int g_enable_tracing = 1;
+static int g_enable_tracing = 1;
+static pthread_key_t g_enable_tracing_key = -1;
+static pthread_once_t tracing_key_once = PTHREAD_ONCE_INIT;
```
drystone: Move this to `get_tracing_enabled()`

common.c
```diff
@@ -177,18 +196,52 @@ trace_dump_data(const void *buf, size_t nbytes)
...
+#else
+	rtr_getpid_t real_getpid = RETRACE_GET_REAL(getpid);
+
+	return syscall(SYS_gettid) == real_getpid();
+#endif
+}
+
 int
 get_tracing_enabled()
 {
```
drystone: Add `static pthread_once_t tracing_key_once = PTHREAD_ONCE_INIT;`

common.c
riataman: I was thinking we weren't supposed to use static inside functions ?

https://github.com/riboseinc/retrace/blob/master/Developers%20Guide.md

At the very bottom: Do not use static storage-class for variables.

common.c
```diff
@@ -52,8 +57,22 @@
...
+ * we will initialized the key g_enable_tracing_key
+ * and will use that to store a per-thread tracing
+ * enabled/disabled setting. I used to declare
+ * g_enable_tracing with the __thread modifier, but in
+ * macOS that will eventually call malloc which would
+ * get us into a loop.
+ *
+ * g_enable_tracing_key logic is backwards, 0 (NULL)
+ * means tracing is enabled. 1 means tracing is disabled.
+ *
```
drystone: Why backwards logic? This is confusing!

common.c
riataman: Having say that I do actually love to use static stuff AND deleraction and initialization mixed in together so I'm really hating all this stuff.

common.c
riataman: Yeah, unfortunately the key is always initialized to NULL and I want this to default to enabled. I guess I could put put some soft if if (isnull) settoenabled() and have enabled be 1 and disabled be 2.

common.c
```diff
@@ -177,18 +196,52 @@ trace_dump_data(const void *buf, size_t nbytes)
...
 {
-	int oldvalue = g_enable_tracing;
+	int oldvalue;
+
+	if (is_main_thread()) {
+		oldvalue = g_enable_tracing;
 
-	g_enable_tracing = enabled;
+		g_enable_tracing = enabled;
+	} else {
```
drystone:
```
{
        oldvalue = get_tracing_enabled();
        pthread_setspecific(g_enable_tracing_key , (void *) (size_t) !enabled);
}
```

common.c
```diff
@@ -177,18 +196,52 @@ trace_dump_data(const void *buf, size_t nbytes)
...
+
+static int
+is_main_thread(void)
+{
+#ifdef __APPLE__
+	return pthread_main_np();
+#else
+	rtr_getpid_t real_getpid = RETRACE_GET_REAL(getpid);
+
+	return syscall(SYS_gettid) == real_getpid();
```
drystone: This could definitely do with braces!

common.c
riataman: nice one :)

common.c
drystone: @riataman sorry for bombarding you with individual comments. I'll try the 'start review' button next time :)
At the very bottom you say ... must have dozed off! OK, leave as is but use get_tracing_enabled() as the only place where the thread local gets initialised if possible.

common.c
drystone: Yes, but can't you set it to whatever value you like in your supplied `initialize_tracing_key`?

riataman: No, remember that is is a per thread storage, pthread_key_create just initializes the key.

Everytime we first call pthread_getspecific from a new thread it'll be NULL.

riataman: I really wanted to use the __thread modifier instead of this, unfortunately __thread variables call malloc in macOS on first read.

drystone: In that case malloc an int if the pointer is NULL and initialise the int to 1. That way you are using the thread local as intended and not keeping coercing an int into the pointer.

riataman: I can't call malloc here, would basically break everything since we haven't disabled traced yet (we are trying to). Passing an int cast to a void * to pthread_setspecific is perfectly acceptable here. If will remove the backward logic and change it to something more clear.

drystone: It's a tricky one. Is real_malloc not an option? it calls other functions I suppose.
Perhaps an enum
```
enum {
        TRACING_ENABLED = 1,
        TRACING_DISABLED
}
```
and use that for the g_tracing_enabled too. That's probably what you had in mind anyway :)

drystone: ... or
```
enum {
        TRACING_ENABLED,
        TRACING_DISABLED
};
```
and then everything is reverse logic but no-one will know!

PR #94, opened 2017-06-13
Some more tidying of the traceing functions and select() fix

drystone:


drystone:
I've added some more to this request: some fixes in char.c and added strchr.

efiop:
@drystone I will introduce new pull request with RFC and style fix and automated check very shortly(1-2 hours). I will have to ask you to wait a little bit with this pull-request, so travis can check it for style too. Sorry for inconvenience!

drystone:
@efiop no problem - it's failing on the mac side - I guess there's no _IO_putc for macs. I'll fix it now.

efiop:
@drystone Thanks for waiting! I've merged my changes, could you please rebase?

drystone:
@efiop, yeah, sure but not until the (UK) morning.

riataman:
@drystone I don't know why you changed this but we need to use this pattern:

 
	int old_tracing_enabled = set_tracing_enabled(0);

        [code we don't want to be traced]

	set_tracing_enabled(old_tracing_enabled);

The reason for this is the following, image this two functions:

do_something ()
{
   set_tracing_enabled(0);
   result = do_one_thing_we_dont_want_traced();
   trace_printf (result);

   // Tracing will be enabled at this point, which we DON'T want
   result = do_another_thing_we_dont_want_traced();
   trace_printf (result);
   set_tracing_enabled(1);
}

trace_printf (args)
{
   set_tracing_enabled(0);
   real_printf (args);
   set_tracing_enabled(1);
}

You will see that in this simple example, tracing will be enabled after the first call to trace_printf, which do_something() doesn't expects, it spects tracing to be disabled for the whole function. Doing the:

old_value = set_tracing_enabled(0);
do_something();
set_tracing_enabled (old_value);

pattern restores the trace value to what it used to be,  which could potentially be disabled, rather than enabling it outright.

riataman:
Also I was thinking instead of buf[i] > 31 && buf[i] < 127  we should use isprint not sure about that one that will depend on the current locale.

At some point it would be nice to have a config to change the dump format. I'm thinking along the lines of:

print only the hex dump
print only ascii
print hexdump + ascii
print everything only as utf8 (useful for web stuff ...).
save the dumps to a separate file (or files, we could create different files based on connection/file).

drystone:
@riataman The places where I removed the `int old_tracing_enabled = set_tracing_enabled(0)` were directly after `if (get_tracing_enabled() == 0) return;` so  what possible values could `old_tracing_enabled` hold (assuming `g_tracing_enabled` is binary?) We know it's a 1 so where's the point in storing it in a variable? If there's something I've missed please let me know - wouldn't be the first time :)

isprint is a good idea, as are the format options ... we should look at hexdump to get some consistency

riataman:
@drystone 

You are 100% correct. old_tracing_enabled  will always be 1 in those cases. I'd still leave the pattern in just to be consistent across the whole code base. The subtlety that in this particular case its always 0 may escape someone reading the code and may copy the code into a different context where its not the case. I don't think we gain in code clarity/speed or anything else by remove it so we might as well leave that in for code homogeneity.

PR #95, opened 2017-06-13
getenv() bugfix and unit tests

drystone:
Fixes getenv() for the case when the environment variable doesn't exist. Also adds some unit tests.


env.c
```diff
@@ -48,8 +48,11 @@ char *RETRACE_IMPLEMENTATION(getenv)(const char *envname)
 {
 	rtr_getenv_t real_getenv = RETRACE_GET_REAL(getenv);
 	char *env = real_getenv(envname);
-	trace_printf(1, "getenv(\"%s\"); [%s]\n", envname, env);
-	return real_getenv(envname);
+	if (env != NULL)
+	    trace_printf(1, "getenv(\"%s\"); [\"%s\"]\n", envname, env);
+	else
+	    trace_printf(1, "getenv(\"%s\"); [NULL]\n", envname, env);
+	return (env);
```
efiop: Just a minor one =): No need to wrap it in braces if it is just a simple variable.

drystone: @efiop - agreed =) but from the [style doc](https://www.freebsd.org/cgi/man.cgi?query=style&sektion=9)

```
Values in return statements should	be enclosed in parentheses.

     Use err(3)	or warn(3), do not roll	your own.

	     if	((four = malloc(sizeof(struct foo))) ==	NULL)
		     err(1, (char *)NULL);
	     if	((six =	(int *)overflow()) == NULL)
		     errx(1, "number overflowed");
	     return (eight);
     }
```

efiop: That's correct, I still confuse Linux and BSD styles sometimes.  Good thing checkpatch.pl isn't confused by it.

PR #96, opened 2017-06-13
retrace: coding style RFC and cleanup

efiop:
This is initial cleanup and RFC to try to unify much of the code.

1) Build retrace with strict CFLAGS, i.e. -Wextra -ansi -pedantic
2) Cleanup and RFC code to make it at least somewhat follow BSD's style(9)
3) Temporarily disable clang-format hook, while I'm working on making it
   suit our needs.
4) Make travis use checkpatch.pl from Linux Kernel to at least check for simple style errors.


efiop:
@riataman The key phrase is "was going". In current state it was an unnecessary typedef that could be avoided. And rtr_config should've had _t suffix to avoid confusion. No offence to you, feel free to introduce what you feel like appropriate in your upcoming patch, I'm open to public discussion.

@drystone I would actually like to see strongly typed structure too. =)

PR #97, opened 2017-06-13
Get this working in FreeBSD

riataman:
I gave it a go and got this to work in FreeBSD 11 (amd64, clang).

We are getting into a mess of #if FreeBSD, #if Apple, #if Linux, hopefully we would be able to test for particular properties in the configure script that is in the work rather than for specific operative systems.


erikbor:
Awesome to see retrace working on FreeBSD

Makefile
```diff
@@ -5,6 +5,8 @@ RM		= rm -f
 
 ifeq ($(OS),Darwin)
 	RETRACE_CFLAGS = $(CFLAGS) -fPIC -D_GNU_SOURCE -Wall
+else ifeq ($(OS),FreeBSD)
```
efiop: No need to retype common CFLAGS over and over again, just move common part out of "if" and append what is needed for each OS.

Makefile
```diff
@@ -15,13 +17,16 @@ ifeq ($(OS),Darwin)
 endif
 
 ifeq ($(OS),Darwin)
-	RETRACE_LDFLAGS = $(LDFLAGS) -dylib
+	RETRACE_LDFLAGS = $(LDFLAGS) -dylib -ldl
```
efiop: Please merge this "if" with the one above. We can easily combine them nicely info one "if".

char.c
```diff
@@ -27,8 +27,16 @@
 #include "file.h"
 #include "char.h"
 
+#if defined(__FreeBSD__)
+#define _DONT_USE_CTYPE_INLINE_
+#include <runetype.h>
+#endif 
 #include <ctype.h>
 
+#if defined(__FreeBSD__)
```
efiop: No need for "defined()" in single-condition if. Just use ifdef like you do in other files.

exec.c
```diff
@@ -336,6 +336,7 @@ RETRACE_REPLACE(execveat)
 int RETRACE_IMPLEMENTATION(fexecve)(int fd, char *const argv[], char *const envp[])
 {
 	int i;
+	rtr_fexecve_t real_fexecve = RETRACE_GET_REAL(fexecve);
```
efiop: Please don't use functions in declarations. Just split it up:
rtr_fexecve_t real_fexecve;

real_fexecve = RETRACE_GET_REAL(fexecve);

trace.c
```diff
@@ -61,7 +65,7 @@ long int RETRACE_IMPLEMENTATION(ptrace)(enum __ptrace_request request, ...)
 		case PT_READ_D:
                         request_str = "PT_READ_D";
                         break;
-		case PT_READ_U:
+		case 3: /* PT_READ_U: */
```
efiop: Does FreeBSD not have PT_READ_U defined? Then just wrap it into #ifndef . No need to hardcode this value for every platform.

riataman: PT_READ_U was removed from FreeBSD.

Historically PT_READ_U has been 3 since ptrace() was first written.

The fact that PT_READ_U is no longer defined in FreeBSD doesn't mean that there aren't binaries with this compiled in. 

I think hardcoding the 3 here is perfectly acceptable instead of creating a #ifdef mess

drystone: I agree with @efiop , #ifdndef this out for FreeBSD, and add a `default:` to catch

riataman:
I updated this, but this is unlikely to still be working on FreeBSD after so many playing around with the code.

ronaldtse:
@riataman it's too bad that Travis does not support FreeBSD (for tests), but apparently some people have been doing exactly that: http://erouault.blogspot.com/2016/09/running-freebsd-in-travis-ci.html

riataman:
That seems like crazy enough to work :) If someone wants to do that be my guest.

I'm thinking we can have a 'dev' branch and use master for stuff when everything compiles builds and runs.

ronaldtse:
Agree @riataman , only working code should be in master. Feel free to create the dev branch!

drystone:
@riataman @ronaldtse  only working code does go into master as it's only merged when the tests pass. Perhaps have Travis for all platforms on master and only the quick (Linux) tests on the dev branch? That way we can merge code quickly into the dev branch and any problems then get flagged up as we create pull requests from dev->master? Or is that what you meant? My issue is with how long it takes for the Mac builds to complete.

riataman:
Well, that's the theory, in reality master is pretty broken right now other than for very simple programs. I have a bunch of fixes that haven't' been merged than bring it up to par.

PR #50, opened 2017-06-08
Optimisations and unit tests for trace_printf and trace_printf_str

drystone:

PR #103, opened 2017-06-14
consolidate trace_printf and trace_printf_str into trace_printf

jjr840430:


common.c
```diff
@@ -65,6 +65,88 @@ int g_enable_tracing = 1;
 struct descriptor_info **g_descriptor_list;
 unsigned int g_descriptor_list_size;
 
+// special strings to be displayed with ANSI colors
```
efiop: Please don't use c++-style comments. Use /* Your comment here */ instead.

common.c
```diff
@@ -65,6 +65,88 @@ int g_enable_tracing = 1;
...
+	{ '\0', NULL }
+};
+
+static void
+trace_printf_str(const char *str)
+{
+	char *str_print;
+
+	// get original function pointers
+	rtr_malloc_t real_malloc = RETRACE_GET_REAL(malloc);
```
efiop: Please don't mix declarations and function calls. I.e. it should look something like:

char *str_print;
rtr_malloc_t real_malloc;
rtr_free_t real_free;
....

real_malloc = RETRACE_GET_REAL(malloc);
real_free = RETRACE_GET_REAL(free);

common.c
jjr840430: Thanks for your advise.

efiop:
Hi @jjr840430 !

Travis found some issues https://travis-ci.org/riboseinc/retrace/jobs/242758802 with your patch. You could test it out locally by running make test to make sure it works for you without errors, most of them are currently about unused variables and similar stuff.

jjr840430:
@efiop Thanks, I'm now checking.

jjr840430:
@efiop Can you check why CI checks has failed? It seems there is no issue from compilation.

efiop:
@jjr840430 It's style check that caught an issue:

WARNING: space prohibited between function name and open parenthesis '('
#50: FILE: test/str.c:35:
+	char *p = strstr (s, "[test\n]");

Please fix it.

efiop:
@jjr840430 Perfect! Thank you.

riataman:
1) This is disabling useful tests in  test/str.c .

2) trace_printf_str is calling strlen directly rathern than real_strlen in one spot.

3) This is removing the [SNIP] functionality which was the entire point of the trace_printf_str function!

jjr840430:
@erikbor I fixed that issue, Can you review and merge?

test/str.c
```diff
@@ -27,11 +27,13 @@
...
 {
-	char *s = "This is a testing string. All work and no play makes Jack a dull boy. Lorem ipsum dolor sit amet, consectetur adipiscing elit";
+	// char *s = "This is a testing string. All work and no play makes Jack a dull boy. Lorem ipsum dolor sit amet, consectetur adipiscing elit";
+	char *s = "This is a testing";
 	char buf[1024];
 	int a = 2;
 
-	char *p = strstr (s, "dull");
+	char *p = strstr(s, "[test\n]");
+#if 0
```
riataman: Why are we removing this tests?

common.c
```diff
@@ -85,66 +178,17 @@ trace_printf(int hdr, const char *fmt, ...)
...
 	set_tracing_enabled(old_tracing_enabled);
-}
-
-void
-trace_printf_str(const char *string)
-{
-	static const char CR[] = VAR "\\r" RST;
-	static const char LF[] = VAR "\\n" RST;
-	static const char TAB[] = VAR "\\t" RST;
-	static const char SNIP[] = "[SNIP]";
```
riataman: The whole point of this function was the [SNIP] stuff, which you didn't carry over to your rewrite.

common.c
```diff
@@ -65,13 +65,106 @@ int g_enable_tracing = 1;
...
+		int idx = 0;
+
+		/* check NULL */
+		if (!colors->color_str)
+			break;
+
+		while (*p != '\0') {
+			if (*p == colors->ch) {
+				char *tmp;
+				size_t size = strlen(str_print) + strlen(colors->color_str) + 1;
```
riataman: We need to be careful to call our replacements directly as we might fall into loops. Call real_strlen instead.

common.c
jjr840430: @riataman Can you advise what's mean of SNIP?

jjr840430:
@drystone The goal of this issue is to consolidate trace_printf() and trace_printf_str().
trace_printf_str() function has only function to display special characters by some colors.
I think it may be consolidated into single trace_printf() function.

jjr840430:
@riataman 
I think [SNIP] is to show whenever logging string has been truncated by limitation of the buffer size in old trace_printf_str() function.
But current trace_printf_str() function allocates buffers by dynamically if it needs.
So I think we don't need that in this issue.

drystone:
@jjr840430 that's correct re SNIP! I'm not sure how you'd consolidate the two without losing functionality. `trace_printf_str` replaces `"\n"` with `"\\n"` which wouldn't be desirable in `trace_printf` unless you wanted never to print a newline :)

jjr840430:
@drystone Can you review that function in this branch?
If you have any opinions, then please let me know.

jjr840430:
@riataman I have new updates. Can you please review the changes?

jjr840430:
@drystone Thanks for your advise.
@erikbor Can you clarify again about consolidation of two functions?

PR #109, opened 2017-06-16
checkpatch: ignore 'return is not a function' error

efiop:
In BSD style it is actually required to use parentheses.

Signed-off-by: Ruslan Kuprieiev <kupruser@gmail.com>


ronaldtse:
LGTM, thanks @efiop !

PR #110, opened 2017-06-16
Tracing and redirection for SSL_get_verify_result

riataman:


drystone:
@riataman is this working on more recent debians? On debian stretch (and therefore probably sid) i get:

```
ssl.c: In function ‘print_ssl_keys’:
ssl.c:44:9: error: dereferencing pointer to incomplete type ‘SSL {aka struct ssl_st}’
      ssl->s3 != NULL &&
```

riataman:
Seems like they made this an opaque object in OpenSSL 1.1.0. Let me see ...

riataman:
Good grief, I'm testing in debian sid now. It has the ssl 1.1 includes, which don't define the SSL type. But the curl (what I'm using to test) it ships is linked vs libssl 1.0  which doesn't have the functions to grab the stuff I need :(

ssl.c
```diff
@@ -123,3 +174,370 @@ int RETRACE_IMPLEMENTATION(SSL_accept)(SSL *ssl)
...
+	long r;
+	SSL *ssl = NULL;
+	char *cmd_str;
+
+	real_BIO_ctrl = RETRACE_GET_REAL(BIO_ctrl);
+
+	r = real_BIO_ctrl(bp, cmd, larg, parg);
+
+	switch (cmd) {
+#ifdef BIO_CTRL_RESET
```
riataman: I hate this. Can someone come up with a more compact pattern?

PR #111, opened 2017-06-16
Added realloc, calloc functions

jjr840430:


riataman:
This basically broke the entire retrace for me ...

jjr840430:
@riataman I'm getting same issue as your case.

dev@dev-virtual-machine:~/work/retrace$ ./retrace.sh ./test/malloc
./test/malloc: symbol lookup error: ./retrace.so: undefined symbol: dlsym

Do you think that the issue is related with adding calloc and realloc?

riataman:
The issue is two fold:

1) We are linking to pthreads
2) We implemented calloc

When we link to pthreads for whatever reason glibc goes into a different path for dlsym which makes it calls calloc, and that point we have falled into a loop.

So it wasn't caught by the tests because the pull requests worked fine on their own separately.

jjr840430:
@riataman I have tested on missing_mem_funcs branch, but it's working well.
It seems there is some bugs on master branch.

I didn't still merged master into missing_mem_funcs branch.
Can you test on missing_mem_funcs branch?

jjr840430:
@riataman I tried comment calloc, realloc functions on master branch, but the result is same.
I think it's not related with those functions.

riataman:
@ jjr840430 I put in a fix. It's not crashing for me anymore.

drystone:
Hi @riataman What was that fix because this commit has broken retrace for me too (on debian testing)? `calloc` seems to be the culprit in my case. I'm getting segfaults for all the tests and I'm assuming it's related to the `dlsym` loop.

riataman:
I haven't figured out why in some cases dlsym calls calloc (and realloc calls getenv). I'm my machine that was happening when I linked retrace to pthread, so I modified the makefile to stop linking to pthread. If that hasn't fixed for you ... I'm investigating different approaches to fix this for good. Apparently there's a barebones __libc_dlsym call in glibc that you can call that doesnt do any locking/allocations that we would use in a constructor to grab the addresses of at least the functions we know dlsym sometimes calls (realloc, getenv).

Other than that maybe do a combined approach of ptrace + our current ld_preload approach.

I have a few ideas that I want to pursue but probably won't have any chance today. 

If anyone else finds a way to fix this for good will have my gratitude. I have spend some time going over the glibc code to see what other approaches are there, go hopefully I can come up with something.

jjr840430:
@riataman @drystone 
This is just Makefile issue.
I fixed and merged into master branch.

jjr840430:
Sorry, it still has segfault issue.
$ LD_PRELOAD=$PWD/retrace.so ls
Segmentation fault (core dumped)

jjr840430:
I disabled calloc(), realloc() functions. These functions are bring segfault issues. I will checkout that reason.

PR #113, opened 2017-06-16
Settings to log to a file and to disable data dumps

riataman:
Still not proposing this for merging. More to get comments from others.


erikbor:
Pablo, I like this. Should we have a switch to disable ANSI colors? Esp when logging to a file we don't need coloring.

PR #121, opened 2017-06-19
added scanf family

maheitsec:


scanf.h
```diff
@@ -0,0 +1,18 @@
+#ifndef __SCANF_H__
```
efiop: It should be __RETRACE_SCANF_H__, just like in other headers.

scanf.h
```diff
@@ -0,0 +1,18 @@
...
+typedef int (*rtr_vfscanf_t)(FILE *stream, const char *format, va_list ap);
+
+RETRACE_DECL(scanf);
+RETRACE_DECL(fscanf);
+RETRACE_DECL(sscanf);
+RETRACE_DECL(vscanf);
+RETRACE_DECL(vsscanf);
+RETRACE_DECL(vfscanf);
+
+#endif  /* __SCANF_H__ */
```
efiop: You should probably add a newline here, both git and checkpatch are not very about having no new line at the end of file.

scanf.h
maheitsec: Fixed it.

PR #126, opened 2017-06-20
Fix deadlocks by using an internal glibc function rathern than dlsym.

riataman:
Restore the memory allocation functions
Fix the data dump in the sock functions


drystone:
Excellent work Pablo, thanks :)

PR #132, opened 2017-06-21
Loads config file only once

drystone:
config now stored in memory. Please review / comment

### DIFF
--- a/common.c
+++ b/common.c
@@ -45,6 +45,7 @@
 
 #include <stdarg.h>
 #include <errno.h>
+#include <sys/queue.h>
 
 #include "common.h"
 #include "str.h"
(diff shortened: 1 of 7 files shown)


common.c
```diff
@@ -314,19 +319,13 @@ get_config_file()
 	rtr_getenv_t real_getenv;
 	int olderrno;
 
-	if (!get_tracing_enabled())
```
riataman: Any reason for removing this? Later in this function we call strcpy and strcat and those are going to show up in the logs causing noise OR loops. Even if those don't show in the log (because they are built in in some platforms)  we are calling stuff like real_fopen which internally could potentially call anything in some platform.

drystone: @riataman thanks for looking this over. `rtr_parse_config` does the `trace_disable()` then calls `get_config()` which in turn calls `get_config_file()`

I'd like to have put an assertion in. Any opinions on using assert()?

riataman: Looks good then. I was trying to follow the call tree but got confused :)

I would certainly use assert, I had an assert in RETRACE_REPLACE at some point to check that we were actually finding the functions, but seems like it never made its way to master.

PR #135, opened 2017-06-23
Add netdb api

jjr840430:


netdb.c
```diff
@@ -0,0 +1,264 @@
...
+
+struct hostent *RETRACE_IMPLEMENTATION(gethostbyaddr)(const void *addr, socklen_t len, int type)
+{
+	rtr_gethostbyaddr_t real_gethostbyaddr;
+	struct hostent *hent;
+
+	real_gethostbyaddr = RETRACE_GET_REAL(gethostbyaddr);
+
+	hent = real_gethostbyaddr(addr, len, type);
+	if (addr) {
```
drystone: Nothing traced if `addr` is `NULL`

netdb.c
```diff
@@ -0,0 +1,264 @@
...
+		char ip_addr[INET6_ADDRSTRLEN];
+
+		if (type == AF_INET) {
+			struct in_addr *in_addr = (struct in_addr *) addr;
+
+			inet_ntop(AF_INET, in_addr, ip_addr, sizeof(ip_addr));
+		} else if (type == AF_INET6) {
+			struct in6_addr *in6_addr = (struct in6_addr *) addr;
+
+			inet_ntop(AF_INET, in6_addr, ip_addr, sizeof(ip_addr));
```
drystone: AF_INET6` ? Do you need `if` `else` here? Would plain `inet_ntop(AF_INET, addr, ip_addr, sizeof(ip_addr));` do?

netdb.c
```diff
@@ -0,0 +1,264 @@
...
+
+void RETRACE_IMPLEMENTATION(sethostent)(int stayopen)
+{
+	rtr_sethostent_t real_sethostent;
+
+	real_sethostent = RETRACE_GET_REAL(sethostent);
+
+	trace_printf(1, "sethostent(%d)\n", stayopen);
+
+	return real_sethostent(stayopen);
```
drystone: Don't need `return` here

netdb.c
```diff
@@ -0,0 +1,264 @@
...
+
+void RETRACE_IMPLEMENTATION(endhostent)(void)
+{
+	rtr_endhostent_t real_endhostent;
+
+	real_endhostent = RETRACE_GET_REAL(endhostent);
+
+	trace_printf(1, "endhostent()\n");
+
+	return real_endhostent();
```
drystone: Don't need `return` here

netdb.c
```diff
@@ -0,0 +1,264 @@
...
+			addr_list = (struct in6_addr **) hent->h_addr_list;
+			for (i = 0; addr_list[i] != NULL; i++) {
+				char ip_addr[INET6_ADDRSTRLEN];
+
+				inet_ntop(hent->h_addrtype, addr_list[i], ip_addr, sizeof(ip_addr));
+				if (i > 0)
+					trace_printf(0, ",");
+				trace_printf(0, "%s", ip_addr);
+			}
+		}
```
drystone: Can you use `inet_ntop(hent->h_addrtype, hent->addr_list[i], ip_addr, sizeof(ip_addr));` so no need for `if` `else`?

netdb.c
```diff
@@ -0,0 +1,264 @@
...
+
+			if (rp->ai_family == AF_INET) {
+				struct sockaddr_in *in_addr = (struct sockaddr_in *)rp->ai_addr;
+
+				inet_ntop(rp->ai_family, &(in_addr->sin_addr), addr, sizeof(addr));
+			} else if (rp->ai_family == AF_INET6) {
+				struct sockaddr_in6 *in_addr = (struct sockaddr_in6 *)rp->ai_addr;
+
+				inet_ntop(rp->ai_family, &(in_addr->sin6_addr), addr, sizeof(addr));
+			}
```
drystone: Same again ... I think `sin_addr` and `sin6_addr` would be in the same position

netdb.c
```diff
@@ -0,0 +1,264 @@
...
+		trace_printf(1, "gethostbyname(\"%s\") = [", name);
+
+		addr_list = (struct in_addr **) hent->h_addr_list;
+		for (i = 0; addr_list[i] != NULL; i++) {
+			char ip_addr[INET6_ADDRSTRLEN];
+
+			inet_ntop(hent->h_addrtype, addr_list[i], ip_addr, sizeof(ip_addr));
+			if (i > 0)
+				trace_printf(0, ",");
+			trace_printf(0, "%s", ip_addr);
```
drystone: Could you use the pattern `trace_printf(0, i > 0 ? ", %s" : "%s", ip_addr)` here and in a few other places. It's less code and loses an unnecessary call to `trace_printf`

jjr840430: @drystone Thanks for your reviews. Will do modify as you mentioned.

PR #136, opened 2017-06-23
new fixup

drystone:
I'm really sorry to do this - another system-wide change!

Inspired by ELF's GOT mechanism, I've reworked the real_ function fixup. At load time the real_ pointers point to a fixup function which when run replaces the real_ pointer from dlsym and then calls it (a little more complicated for the vararg functions).

So there is no longer a RETRACE_GET_REAL wrapper, you are free to use real_{func} whereever you need.

Tests pass I think, but only on linux so far ... I've updated the apple/bsd code but it's as yet untested.

### DIFF
--- a/char.c
+++ b/char.c
@@ -41,12 +41,7 @@ static void
 trace_putc(const char *name, int c, FILE* stream)
 {
 	static char specials[] = "\nn\rr\tt";
-	rtr_fileno_t real_fileno;
-	rtr_strchr_t real_strchr;
 	char *p;
-
-	real_fileno = RETRACE_GET_REAL(fileno);
-	real_strchr = RETRACE_GET_REAL(strchr);
 
 	p = c == '\0' ? " 0" : real_strchr(specials, c);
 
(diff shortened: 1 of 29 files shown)


efiop:
I really like the idea, but it seems like the macros got less intuitive:

-RETRACE_REPLACE(putenv)		 +RETRACE_REPLACE(putenv, int, (char *string), (string))

 Maybe there is a way to use defined rtr_func_t types or something to make it better?

drystone:
@efiop I know it's ugly, but I can't think of a better way other than to preprocess (linke  QT an the moc compiler). The problem is that the fixup function needs to chain the real function otherwise we're back to needing two calls and we lose the convenience of being able to use `real_funcs` without assigning them every time we use them. The `RETRACE_REPLACE_V` is even uglier as you can't chain vararg functions without employing the stdarg macros. On balance though, I think this is a worthwhile change.

efiop:
@drystone I agree, this looks like the best way possible.

ronaldtse:
Looks like this can be merged now!

PR #137, opened 2017-06-24
added netdb functions

jjr840430:


test/cmockatest.c
```diff
@@ -352,6 +353,31 @@ RTR_TEST_END
 RTR_TEST_START(dirfd)
 RTR_TEST_END
 
+/* netdb functions test */
```
efiop: Can we also have unit tests here? I see that you've added standalone binaries, but they don't really test anything always returning 0.

jjr840430: @efiop OK, I will add testing functions.

netdb.c
```diff
@@ -0,0 +1,241 @@
...
+			}
+		} else if (af == AF_INET6) {
+			struct in6_addr **addr_list;
+
+			addr_list = (struct in6_addr **) hent->h_addr_list;
+			for (i = 0; addr_list[i] != NULL; i++) {
+				inet_ntop(hent->h_addrtype, addr_list[i], ip_addr, sizeof(ip_addr));
+				trace_printf(0, i > 0 ? ",%s" : "%s", ip_addr);
+			}
+		}
```
drystone: @jjr840430 as per previous comment, this if/else is  unnecessary. `addr_list` is an array of pointers, `inet_ntop` does the cast to `in_addr *` or `in6_addr *` depending on `hent->addrtype` so you don't need to:

`inet_ntop(hent->h_addrtype, hent->h_addr_list[i], ip_addr, sizeof(ip_addr));`

test/cmockatest.c
```diff
@@ -352,6 +353,62 @@ RTR_TEST_END
...
+RTR_TEST_START(getaddrinfo)
+	struct addrinfo hints, *result;
+
+	memset(&hints, 0, sizeof(struct addrinfo));
+	hints.ai_family = AF_UNSPEC;
+	hints.ai_socktype = SOCK_DGRAM;
+	hints.ai_flags = 0;
+	hints.ai_protocol = 0;
+
+	result = rtr_getaddrinfo("www.google.com", NULL, &hints, &result);
```
efiop: You should use assert_* macros [1] here(and in similar cases) to check if result is appropriate. I.e.:

assert_int_equal(result, 0);

Btw, 'result' should probably be int, not struct addrinfo *.

https://api.cmocka.org/group__cmocka__asserts.html

netdb.c
jjr840430: @drystone Thanks for your review. It's really so.

test/cmockatest.c
jjr840430: @efiop It was fault on testing code. Thanks.
But I think it's not need to use assert_int_equal. Because result may be NULL.
Also it should be checked for NULL before free

jjr840430:
@efiop Can you please review for new commits?

drystone:
@jjr840430 thanks for this, all good stuff! I've rebased and all tests passing. The cmocka tests could do with some fleshing out - all they do is test that the functions exist and run but don't test return values. If you've the inclination, perhaps we'd benefit from more thorough unit tests. Merging ...

PR #143, opened 2017-06-26
Structured data

riataman:
Mode all the functions away from printf and into a struct that describes what to print.

This sets the foundation for:

1) UI
2) Redirect in all functions and parameters.
3) Standardized printing


riataman:
This is a huge one that basically touches everything.

ronaldtse:
@riataman apologies this must have conflicted with the just merged #136 ...

riataman:
No problem I will try to bring it up to date with master later tonight.

riataman:
This was a big merge and some stuff are still missing or broken. I'll finish bringing it up to shape tomorrow.

drystone:
@riataman this looks really good! I can't help thinking that some form of automatic function generation would be a good idea, could save a lot of time and reduce the possibility of typos in what is essentially boilerplate. We've the possibility of huge `#define`s or perhaps, as I mentioned briefly to @efiop , we could write some form of preprocessor that examines the function prototypes and generates all the code for interposing/fixup and the data capture. Something like 
```%.c : %.rtr```
    rtrprep $^ $@
```
where `.rtr` files are C source with some special preprocessing instructions ...

riataman:
@drystone not sure about the parameter list being const, I could potentially change some parameters after some point for some events.

For stuff like utsname I did a custom parameter type, which I don't 100% liked. I wonder if we could potentially just describe the whole struct rathern than special case it (could be a pain for very complex structs). Specially for one-use stuff like utsname which is only going to be used in one spot.

The idea is that you can put the data into a network/pipe whatever, but I'm thinking that some degree of serialization and deserialization is going to be needed (hopefully not that much) as this is not a fixed size struct and consists mostly of pointers (I don't know how we would make this work for pointers).

For files I'm thinking about doing a different event:  EVENT_TYPE_BEFORE_OPEN_FILE, so that a potentially gui could stop the program and to an interactive redirect of the file no mater exactly what functions is trying to do the open. Same with connect() I'm thinking EVENT_TYPE_BEFORE_NETWORK_CONNECTION. Something along those lines. I guess a PARAMETER_TYPE_FILENAME still would make sense in combination with the above.

I'm not going to have time to work on this again for about 5hrs, so if you want to work on this in the meanwhile feel free to do so, at this point the items I'm missing are:

1) Make sure this passes all the tests
2) Fix whatever formatting the script complains about.
3) Make sure this work at least in macOS as I haven't tested anywhere else.
4) I think there are like 2 or 3 functions that I didn't convert and need a bit of thought on how we want to handle them. 

So, feel free to work on this (but let me know I guess so we don't overlap efforts), my goal is to get it into shape to get it merged and then we can work on all the ramifications that this allows, like serializing the data into the network/pipe, better printing colors/format, work on redirection of all parameters and return values, etc.

drystone:
@riataman one more thing, I don't know why but I feel event_type could be a parameter to retrace_event and not a member of the struct

riataman:
@ drystone hahaha, I actually started feeling the same way but was halfway down the rabbit hole with this. Feel free to change it!

riataman:
I think we can merge this now. There are some functions missing and some stuff could certainly be improved but at this point I think there are no obvious regressions and is a good foundation to keep improving it.

Also its increasingly a pain in the arse to maintain and merge stuff into it :)

@drystone  what do you think?

drystone:
@riataman as you know I'm not entirely happy with the extent of this change. I think formalising the trace data/format is an excellent idea and this part of the patch is of real benefit. But you are also introducing a generic mechanism for modifying parameters before `real_func()` and/or the return value. This would be fine if we were using automatically generated shims for the overridden functions. In which case our `malloc.c` would simply contain hooks for pre and post modification functions. As it stands though, our `malloc.c` is a hand coded shim that should be as programmer friendly as possible.

I'd prefer to see separation of tracing and parameter modification and a more explicit code style for parameter modification. It's not nice to have variables updated as a side effect of calling a function and such side effects are a common cause of problems.

For an (unrealistic) example, in your scheme:
```
void *malloc(size_t size)
{
    int r;
    struct rtr_event_info ei;
    ei.params = {&size};  /* all cast to void ** */
    retrace_event(&ei); /* may modify size as a side effect */
                                   /* by the malloc-agnostic modification function */
                                   /* and who does freeing of any allocated buffers for modified data? */
    size = my_malloc_mod(); /* loses modification but an easy mistake to make */
    r = real_malloc(size);
    retrace_event(&ei); /* same again - r may be modified as a side effect */
    r = malloc_fuzz();  /* and modification lost ... */
    return r;
}
```
I think it would be better to have the more explicit
```
void *malloc(size_t size)
{
    struct rtr_call_info ci, mci;
    ci.params[0].intval = size;  /* typed */
    rtr_log_call(BEFORE, &ci);
    if (rtr_redirect_args(&ci, &mci) /* no voodoo */
        size = mci.params[0].intval;
    else
        size = my_malloc_size;
    ci.retval.pvoid = real_malloc(size);
    rtr_log_call(AFTER, &ci);
    return ci.retval.pviod;
}
```
Even this has problems as the before and after logging may get separated in the output in multithreaded programs.

drystone:
Couldn't we have something simpler along the lines of
```
rtr_log(const struct rtr_func_decription func, /* types, parameter names, return type etc */
             union rtr_arg *preargs, /* args as caller intended */
             union rtr_arg *inargs,   /* rtr modified args before real_func */
             union rtr_arg *outargs, /* retval and args after real_func */
             union rtr_arg *postargs, /* rtr modified retval and args */
             int errno);
```
to be called after `real_func` so we get consistent logging. Redirections/fuzzing can then be handled specifically per function using helper functions in `common.c` where appropriate?

riataman:
Well your approach start getting verbose with functions with more parameters, also need to include the return value. Also starts loosing its explicity unless you want to make it really long.

```
void *fopen(const char *file, const char *mode)
{
    unsigned int parameter_types[] = {PARAMETER_TYPE_STRING, PARAMETER_TYPE_STRING, PARAMETER_TYPE_END};  // you still need this here, the union won't magically tell you what type it has.
    struct rtr_call_info ci, mci;
    ci.params[0].charval = file; 
    ci.params[1].charval =mode;

    rtr_log_call(BEFORE, &ci);

    if (rtr_redirect_args(&ci, &mci) {
        // This this really explicit?? since one or the other might have changed
        // need to check each one to make it really, really explicit?
        (char *) file = mci.params[0].charval; // need to ignore the const
        (char *) mode = mci.params[1].charval;        // need to ignore the const
    }
    ci.retval.pvoid = real_malloc(size);

    rtr_log_call(AFTER, &ci);

    // We need flags inside rtr_call_info to telll wherer or not we need to free()
    // file or mode here. 
    if (rtr_redirect_return_val(&ci, &mci) {
        ci.retval.pvoid = mci.retval.pvoid;
    }

    return ci.retval.pviod;
}
```

Also your proposed change:

```
rtr_log(const struct rtr_func_decription func, /* types, parameter names, return type etc */
             union rtr_arg *preargs, /* args as caller intended */
             union rtr_arg *inargs,   /* rtr modified args before real_func */
             union rtr_arg *outargs, /* retval and args after real_func */
             union rtr_arg *postargs, /* rtr modified retval and args */
             int errno);
```

Is arguing to basically break down the struct that I have to function parameters. I'd rather keep the struct so I can add stuff later on without breaking every single call.

If you want to tell the caller that the values may change you can just rename the event function to:

retrace_log_and_redirect_before();
retrace_log_and_redirect_after() ;

And forget about events, and handle stuff like opening a file inside the struct with flags.

I don't think unions over void* is giving us any advantage here, you still need to basically tell the function what value it has so type safety is an illusion, if you set the wrong type in the array it will still behave badly. Explicitly setting the redirect inside the called function just makes it more verbose with any obvious gains and starts getting murky when you have more than one parameter.

If we want to protect stuff from changing under the roof. Like in connect() we can add a PARAMETER_FLAG_DONT_REDIRECT so that the redirection code will ignore that parameter.

riataman:
Also honestly most functions could automatically generated with just a description file, with just a few special cases.

drystone:
@riataman thanks for your comments.

I don't find your `fopen` particularly verbose. I daresay even more verbose in reality once a static function description is added.

Agreed the imaginary `rtr_get_redirect` would ideally be called for each argument. And would be documented to say no need to free strings as they are already in the global config struct. Similarly, any other modification functions would be documented. In your scheme, the modification happens behind the scenes and so we've no idea what to do about freeing. You'll have to create a free function which will potentially blow or leak if someone reassigns a pointer (whether by accident of intentionally).

I also agree that the `union` is fairly pointless as far as type safety goes but for readability it works and as it can contain structs `{buffer, length}` pairs can be stored in a single entry.

The BEFORE and AFTER calls will separate the lines of output. This can all be done in an AFTER call having captured all the necessary information (with the exception of exec and exit). 

And finally, I agree that many shims can be automatically generated. I tried this [a while back](https://github.com/riboseinc/retrace/pull/34). If that is the case, what's the problem with having verbose auto-generated code?

This is a shame because currently we're at a stage where a universal logging format is going to be really useful. We can add configuration things like `showargnames`, `showstructmembernames` or  `logfailures`. We can sort/group/filter reliably with a strict format and marshal the data, nearly as is, into other processes. However I'm not sure we're really ready for a universal redirection mechanism.

riataman:
I always do my designs thinking in what features I want first. This is what I want to support:

- Universal logging: As you say, disable showing parameters, disable showing any function, just show a particular function/functions, show the calls in a tree like view, disable dumps, just follow a single file descriptor, etc.
- Universal redirection: I want to be able to modify any parameter AND any return value. The config file in reality is a very poor example of this usage. I want a powerful interactive GUI where I can basically edit all memory before and after calls. THIS is my end goal with the redirection stuff. I want to see a write() call and I want to see the buffer that is going to be written in the GUI and I want to edit it directly like if I was in an hex editor and have it be feed back into the program transparently.  Same with read, I want to be able to see what was read and edit it on the fly before the program being traced ever see it.

That last part is what I'm more exited about.

Anyway, I think we have reached a dead end with this discussion and its boiling down to personal preference. I'd rather merge what we have and start working on the actual features this will allow and then AS THE FEATURES WE WANT REQUIRE we start tweaking it.

Otherwise we either have everything else stopped or we keep merging stuff in master that uses the old style printf() stuff that is just creating double work down the file.

I agree I don't have the best implementation right now, but I would rather work on useful features, rather than endlessly tweaking infrastructure. Ultimately what is going to attract users are our killer features. I would rather write a 80% solution and then when I'm really have some useful code on top of my infrastructure and I really understand how its being used and I understand the problem more, improve it.

ronaldtse:
This is becoming a very long discussion, but I have the following thoughts:

* For thread tracing, obviously we should trace all threads. We can print stuff out using a queue and an extra "retrace thread". We could start the thread since we trace pthread, and in the future we could trace boost too. This will give us sane output.

* The config file is limiting, and we should have a better alternative. A GUI would be good to develop a binary wrapper, for the sake of it, a custom Firefox. Even when we have a GUI, it should probably only be run in interactive mode for development because you want it to "replay", right? Then it needs to store these complex instructions somewhere. This I believe should be the API / driver. @riataman are you thinking of something like this?

* For the "killer app question" that has come up so many times, I was talking with @erikbor today about overriding / fuzzing network clients / servers. I think @drystone is considering the same with piping network data. 

* It's probably the time for automatically generated shims since @drystone 's attempt and today's mentioning?

riataman:
I got rid of the event stuff (at least in the public interface). and changed it to this:

```
        retrace_log_and_redirect_before(&event_info);

        ret = real_write(fd, buf, nbytes);

        retrace_log_and_redirect_after(&event_info);

```

At the end of the day I don't think we need more than two calls, we can encode everything that is going on inside the struct.

riataman:
Sorry, I missed the thread argument.

I'm not married with login the before and after all. I tough it could be useful to see what calls does a function does internally. I'd actually would like to show this more in a tree like manner. I mostly put that in as a test that I was getting things right.

I just went ahead and disabled the double login of calls. Would need some flag for exit and exec to force login of before in those.

riataman:
@ronaldtse:

I'm thinking on some sort of libretrace which can read from a socket/pipe/shared memory all the tracing info and can display it in realtime, stop the program as needed and modify parameter, return values, random memory, etc.

A GUI will use this libretrace and display and enable funcionality.

ronaldtse:
@riataman `libretrace` will be interesting.

I think if we're going to have a GUI it should be a networked one (a "retrace server") for the benefit of a cross platform UI and the remote interactive functionality (as you mentioned).

@drystone too mentioned the server approach before.

riataman:
@ronaldtse I'm eyeing a network gui too.

I actually implemented a very rought proof of concept here: https://github.com/riataman/retrace/commit/23145e71b609f0ee8e037c00525ccca223abeea7

That branch basically dumped the output of trace_printf to the network. Worked pretty well.

drystone:
@riataman  @ronaldtse I think we're at a crossroads. We either continue to hand craft some or all of the shims, in which case I'm not happy with this patch. I think `free`ing the monster `struct` is going to cause problems. We'll end up having flags in the `struct` to determine whether memory should be `free`d or not. Or the alternative is to generate all the shims and pipe all information to.from wrapper programs. In which case it doesn't matter what mechanism is used assuming it works reliably. Once we've gone down this route, retrace.so is essentially done apart from describing new functions as we add them. Without doubt I favour this second approach. retrace.so will become a clean and simple library with minimal impact on the program being traced. All the formatting/logging/configuration/special cases can be done in a real wrapper program with a `main()`

We will need the following:

- An IPC mechanism. A named pipe would probably be best here, created by the client and picked up as an environment variable by retrace.so. If a pipe is too slow we can look at shared memory.
- A protocol. Raw data ideally. Whatever we do we should include a protocol version number as it's sure to change :)
- Marshall/unmarshall functions
- A shim generator to create headers that can be compiled into the client to aid marshalling and the function definitions for retrace.so

Then we can make the retrace.so shims do something like the following:
```
reredirect_args = ipc_preredirect(function_id, pid, original_args);
copy_args(real_args, preredirect_args);
retval = real_func(real_args);
copy_args(return_args, retval, errno, real_args); 
postredirect_args = ipc_postredirect(function_id, pid, return_args);
ipc_log(function_id, pid, original_args, preredirect_args, return_args, postredirect_args);
errno = postredirect_args.errno;
return postredirect_args.retval;
```
Does that sound like a reasonable approach?

drystone:
Perhaps we should create a dev branch to get this working with generated shims and a wrapper that can pass the test suite before committing to master.

riataman:
@drystone:

I think we are over designing this upfront without the knowledge of how what we want to do is going to turn out.

Can we merge the current code, work on the protocol and other user visible features and once we have seen what we really need for a protocol we can make the changes we need? Instead of changing this ad nauseam like the get_real code without really introducing any user visible benefits?

Same with the universal redirecting, do we want to manually keep redirecting every single function OR we can merge this, work on the global redirect and if it turns out to be a nightmare we can backtrack and adjust?

Any other design we can come up with is going to have the same potential for upsides/downsides, but this code is here now and is the best we have. We can go back and spend another week in an alternative design which may or may not be better. We are basically back at square one.

I'd rather work on the features we want and as we work on those we can improve the infrastructure.

At this point I'm not even arguing in favor of my code but move on and not be stuck.

erikbor:
I want to suggest we ensure this patch passes all the checks and builds properly and then we merge it. That way we don't have a stopper and we can continue working on adding new functions and improve the infrastructure.

Would the use of ptrace for tracing (and thus having a wrapper program for that) be a solution of some of the issues that are lurking on the horizon?

ronaldtse:
@riataman I think we are all in favor of moving forward.

* An automated shims approach is definitely preferred but we should also account for manual overrides, we will want more detail / action in certain functions than others.

* Explicit arguments vs one struct as argument "argument" both have pros and cons. @erikbor 's point of ptrace is the possibility of using some ptrace struct instead? The key here is free function to clean the event struct so we don't leak stuff

Let's try to fix up this PR and merge it first. While this conversation is a bit heated we do have a list of things to do moving forward:

* An IPC mechanism
* A protocol
* Marshall/unmarshall functions
* Shim generator
* Networked retrace / GUI
* Thread support
* ptrace

PR #150, opened 2017-06-27
-f option overide RETRACE_CONFIG

catap:


riataman:
Merged, thanks.

catap:
@wizzard but with your way `retrace -f /path/to/config ...` doesn't work on OS X that I'm using.

wizzard:
@catap the current code causes test apps crash on MacOS when `retrace.conf.example` file is used.
I disabled conf file, please see `retrace.in` file "XXX" comment.

riataman:
@wizzard the tests where crashing because the config file is specifying malloc fuzzing which macOS doesn't like since it uses it internally. It was working before because the -f option was basically no working in macOS. The solution is not to break -f again, but disable the malloc fuzzying option.

wizzard:
Ok, so then we need 2 different config files for our test apps with and w/o malloc fuzzing?

riataman:
Just comment the malloc fuzzing option for now. Not sure what would it take to make it work in macOS. Maybe delay the fuzzing until main has already run somehow?

You can open a new issue for that.

catap:
@riataman looks like OSX has a bug inside DYLD and it never check that malloc may return NULL :)

So, for test proposal it's better to comment, yes.

riataman:
Yeah, what a pain :) in Linux glibc almost never uses the public facing functions internally which is great for retrace as we avoid noise and this kind of bugs. 

On macOS they do use them and its a pain in the arse since there's a lot of noise even before main() is called would need to investigate if something can be done at some point about that.

PR #154, opened 2017-06-28
Added mmap, munmap, brk, sbrk functions

jjr840430:


drystone:
@jjr840430 as you are modifying the random stuff, could you have a look at [deterministic seeding](https://github.com/riboseinc/rnp/pull/255)?

malloc.h
```diff
@@ -6,9 +6,30 @@ typedef void (*rtr_free_t)(void *mem);
...
+
 RETRACE_DECL(malloc);
 RETRACE_DECL(free);
 RETRACE_DECL(calloc);
 RETRACE_DECL(realloc);
 
+RETRACE_DECL(mmap);
+RETRACE_DECL(munmap);
+
+#ifndef __APPLE__
```
catap: Why? macOS has `brk` and `sbrk`

jjr840430: @catap Those functions were deprecated on MacOS, So it was failed to pass Travis CI checking.

catap: wow

ronaldtse: @jjr840430 could you help add a comment there that they are deprecated on macOS, and this line is necessary to pass? Thanks!

ronaldtse:
Looks good @jjr840430 , as @drystone mentioned could you look at #156 / and deterministic seeding mentioned in #255 ?

jjr840430:
@ronaldtse Thanks for your reviews, I will look issues that you mentioned

jjr840430:
@ronaldtse @catap Please review last commits.

You can test with the following configuration line.
memoryfuzzing,0.05,1498729252,1
> 0.05: fail rate
> 1498729252: seed value, if 0, then generated automatically
> 1: enable only fuzzing logging, 0: enable all loggings

catap:
@jjr840430 I think is better to remove logging option from `memoryfuzzing` because it may confuse and also that logic will be if we add `iofuzzing`?

jjr840430:
@catap That option is for only memory fuzzing, so it's not affect to iofuzzing.
Do you have any suggestion to enable/disable verbose logging?

catap:
I have one idea. We should implement a filter that can enable/disable retracing/logging of any call.

Something like this: `retrace -f /path/to/config -e malloc,!free ./a.out`

Also from my side looks useful create a group to this conditions (likes IO, network, memory, etc).

Something like this: `retrace -f /path/to/config -e malloc,!free,!net ./a.out`

But it looks like as another issue/task ;) and my point here — let keep it simple right now.

jjr840430:
@catap , OK, agreed, I will push modification soon.

jjr840430:
@catap Removed logging option, so configuration line may be:
 memoryfuzzing,0.05,1498729252

catap:
@jjr840430 I have one more idea: add optional parameter to print stacktrace when fuzzing works.

What do you think?

jjr840430:
@catap sorry, I didn't fully understand, Can you give some examples?

catap:
@jjr840430 let me make small example that I mean:
```C
if (rtr_get_config_single("memoryfuzzing", ARGUMENT_TYPE_DOUBLE, ARGUMENT_TYPE_END, &fail_chance) &&
		rtr_get_fuzzing_flag(fail_chance)) {
		trace_printf(1, "malloc(%d); [redirecting: NULL]\n", bytes);
		
    void* callstack[128];
    int i, frames = backtrace(callstack, 128);
    char** strs = backtrace_symbols(callstack, frames);
    if (strs != NULL) {
        for (i = 0; i < frames; ++i) {
            trace_printf(1, "%s\n", strs[i]);
        }
        real_free(strs);
    }
    return NULL;
}
```

this code will produce a log file likes:
```
(69698) malloc(64); [redirecting: NULL]
(69698) 0   libretrace.dylib                    0x000000010009c3b1 retrace_impl_malloc + 145
(69698) 1   libc++abi.dylib                     0x00007fff97ba4e2e _Znwm + 30
(69698) 2   libc++.1.dylib                      0x00007fff97b673f0 _ZNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEE6__initEPKcm + 66
(69698) 3   libbotan-2.0.dylib                  0x0000000100240d8c _ZN5Botan8EC_GroupC2ERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 1004
(69698) 4   libbotan-2.0.dylib                  0x000000010017d02a botan_privkey_create_ecdsa + 138
(69698) 5   librnp.0.dylib                      0x00000001000b8815 pgp_ecdsa_genkeypair + 293
(69698) 6   librnp.0.dylib                      0x00000001000b72f7 pgp_generate_keypair + 391
(69698) 7   rnp_tests                           0x0000000100002c71 ecdsa_signverify_success + 273
(69698) 8   rnp_tests                           0x0000000100007e64 main + 612
(69698) 9   libdyld.dylib                       0x00007fff98f98235 start + 1
(69698) 10  ???                                 0x0000000000000001 0x0 + 1
```

os, as result we can see the last stacktrace before crash.

jjr840430:
Wow, that is very wonderful, OK, will do so.

catap:
@jjr840430 I'm thinking may be it should be as separated feature and you can enable it for all retraced calls?

Because right now I see `free(0x0)` at my log and it is very intresting to see the stacktrace.

catap:
@ronaldtse what do you think about this?

jjr840430:
Yes, I think it should be separated feature.

riataman:
@jjr840430 I would also add the random seed as a separate config so we can use it for other fuzzing settings.

jjr840430:
@riataman that's great, thanks.

riataman:
I'm starting to work on iofuzzing, so would be great to have a common function so that everybody can use it:

static int g_init_rand = 0;

retrace_init_random()
{
                if (!init_rand) {
                    unsigned int random_seed;

        if (rtr_get_config_single("fuzzingrandomseed", ARGUMENT_TYPE_UINT, ARGUMENT_TYPE_END, &random_seed)) {
                long int random_value;


                        srand (random_seed);
                        init_rand = 1;
                }
                else
                         srand (time(NULL));


                random_value = rand();
}

ungisned int retrace_get_random()
{
           /* Lets do this so we can add our own random generator later, as not to disturb
               the running program random state */
           return rand();
}

```

**Comment by ronaldtse** [comment]
@catap I think it's the right thing to do!

@jjr840430 @riataman agree to make sure retrace's usage of random numbers does not affect the original executable's usage of it.

**Comment by catap** [comment]
@ronaldtse and today I realised that the best way is calling `print stacktrace` as separated log-level and it is absolutelty irrelevant with fuzzing :)

**Comment by jjr840430** [comment]
@catap Almost done, I did spend much times to solve conflicts while merging master into this branch.
I will update soon.

**Comment by jjr840430** [comment]
@catap I pushed, Please review, Thanks.

File: common.c [hunk 2]
```diff
@@ -1101,6 +1102,27 @@ trace_mode(mode_t mode, char *p)
...
+/* printf backtrace callback */
+void trace_printf_backtrace(void)
+{
+	void *callstack[128];
+	int old_trace_state;
+
+	old_trace_state = trace_disable();
+
+	int i, frames = backtrace(callstack, 128);
+	char **strs = backtrace_symbols(callstack, frames);
```
catap: Here I see one challenge: `backtrace_symbols` is using inside malloc for example at [gnu libc](https://github.com/lattera/glibc/blob/master/debug/backtracesyms.c#L71).

And it's should create an infinity loop of printing stacktrace.

jjr840430:
Currently it disabled tracing before calling backtrace, so it would not be in infinite loop.
It's working with testing codes.

common.c
jjr840430: Currently it disabled tracing before calling backtrace, so it would not be in infinite loop.
It's working with testing codes.
Are you getting infinite looping issue?

catap: Sorry, I missed it. Confirm that it hasn't got an infinite loop.

jjr840430: OK, I added the line to check tracing status before calling backtrace() function.

catap: Thanks!

jjr840430: @catap Fixed SSL issue on osx,  but it seems already fixed by you.

catap:
@jjr840430 may you just remove last commit with SSL fixing and use `git push -f`? ;)

jjr840430:
@catap Done.

catap:
LGTM!

Issue #170, opened 2017-06-30
Can't build last build on macOS


catap:
@riataman, looks like your commit https://github.com/riboseinc/retrace/commit/4a23dda3c4636f084c4e89038f8667514a492b25 crash building on macOS.

Since this commit I have:
```
  CC       libretrace_la-trace.lo
trace.c:223:1: warning: incompatible pointer types initializing 'rtr_ptrace_t' (aka 'long (*)(int, ...)') with an expression of type
      'int (int, pid_t, caddr_t, int)' (aka 'int (int, int, char *, int)') [-Wincompatible-pointer-types]
RETRACE_REPLACE_V(ptrace, long int, (int request, ...), request, ptrace_v,
^                 ~~~~~~
./common.h:105:16: note: expanded from macro 'RETRACE_REPLACE_V'
rtr_##func##_t real_##func = func;
               ^             ~~~~
<scratch space>:38:1: note: expanded from here
real_ptrace
^
1 warning generated.
  CC       libretrace_la-write.lo
  CCLD     libretrace.la
Undefined symbols for architecture x86_64:
  "_print_ssl_keys", referenced from:
      _retrace_dump_parameter in libretrace_la-common.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libretrace.la] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

catap:
I found and fixed one issue https://github.com/riboseinc/retrace/commit/22b8cde502690b16f727c6c82948cdb295e526d8

But it still doesn't enable `ssl.c` and `ssl.h` to build.

catap:
and it's fixed at #171

PR #171, opened 2017-06-30
Fix autotools for #170

catap:


catap:
@jjr840430 yep and if you can fix it right now feel free to do it because I'm not near my laptop.

catap:
@jjr840430 done at https://github.com/riboseinc/retrace/pull/171/commits/eda11978383b65a66476d832ee1608a049b1d151

PR #172, opened 2017-06-30
Incomplete IO Mode

riataman:
Move the seed into its setting so other code can use it
Make the seed printing part of the event info struct
Make incompleteio use the new code


riataman:
@jjr840430 ta a look at this one. I moved the seed setting to its own setting so it can be shared by other modes. Also added some fields to add extra info to the struct printing.

jjr840430:
@riataman Great, I agreed. Thanks.

PR #182, opened 2017-07-03
rtrace.in: remove -e from echo because builtin echo of /bin/sh does not support that, use a for loop and eval to simplify things

erikbor:
'-e' does not work with /bin/sh builtin
use a for loop in get_lib_file() to replace repetitive checks
use eval to simplify macos/others environment variables


riataman:
Looks good to me.

PR #186, opened 2017-07-04
Added memcpy, memmove, bcopy, memccpy

jjr840430:


jjr840430:
Thanks @drystone I forgot it :+1:

jjr840430:
@catap I replaced memcpy() function into memmove() when memory overlapping is occurred. Please review memcpy() function.

malloc.c
```diff
@@ -179,6 +179,120 @@ void *RETRACE_IMPLEMENTATION(realloc)(void *ptr, size_t size)
...
+
+	retrace_log_and_redirect_before(&event_info);
+
+	/* check overlapped memory copying */
+	if (abs(dest - src) < n) {
+		use_memmove = 1;
+		event_info.extra_info = "redirected : memmove() (overlapped)";
+	}
+
+	if (!use_memmove)
```
catap: I think it is a bad idea. Show that 'memcpy' has overlapping is good point, but change the syscall... It change behaviour of program.

My point here that trace utilite should show the bug but don't try to solve it.

jjr840430: @catap OK, then I will show error message at this time. What do you think about redirecting NULL at overlapped event?

catap: @jjr840430 I think that we should show 'Overlapp memory happened on memcpy, it may create a bug! please read the man'.

And call `real_memcpy` as it requestd.

jjr840430: @catap I fixed it.

catap: @jjr840430 thanks!

PR #190, opened 2017-07-04
Do not open/close logfile each trace_printf and made flush optional

catap:


riataman:
I really hate this script of ours ...

ERROR: do not initialise statics to NULL
#30: FILE: common.c:512:
+	static FILE *output_file = NULL;

catap:
@riataman I absolutely agree!

riataman:
Fixes tests and made our mighty script happy.

ronaldtse:
"Our" script has been a source of contention, but this is actually not "our" script, it's the `checkpatch.pl` script to enforce best practices from the Linux kernel project. We are merely using it for code style checking as you can see in `ci/install.sh`...

PR #194, opened 2017-07-04
Show real arguments at call + fix for OpenSSL

catap:


riataman:
Looks good to me, what is the OpenSSL fix?

PR #195, opened 2017-07-05
Trace children

riataman:
@catap This one is for you.

Its barely tested in Linux and not at all in macOS.

Seems to be working at first glance. Will test it a little bit more tomorrow, specially in macOS.

### DIFF
--- a/exec.c
+++ b/exec.c
@@ -30,8 +30,8 @@
 #include <unistd.h>
 
 
-/* Duplicates a string array adding an extra spots,
- * return number of spaces in extra_space
+/* Duplicates a string array adding extra spaces,
+ * returns the new array that must be free'd
  */
 static char **
 duplicate_string_array(const char **array, int extra_space, int *count)
@@ -63,7 +63,7 @@ static char*
 find_environment_var(const char *var)
 {
 	char **s;
-	int len = real_strlen(var);
+	int len;
 
 	len = real_strlen(var);
 
--- a/exec.c
+++ b/exec.c
@@ -26,7 +26,86 @@
 #include "common.h"
 #include "exec.h"
 #include "malloc.h"
+#include "str.h"
 #include <unistd.h>
+
+
+/* Duplicates a string array adding an extra spots,
+ * return number of spaces in extra_space
+ */
+static char **
+duplicate_string_array(const char **array, int extra_space, int *count)
+{
+	const char **s;
+	char **ret;
+
+	*count = extra_space + 1;
+
+	if (array)
+		for (s = array; *s; s++)
+			(*count)++;
+
+
+	/* relying in the fact that calloc initializes memory to zero */
+	ret = (char **) real_calloc(*count, sizeof(char *));
+
+	if (ret) {
+		int i;
+
+		for (i = 0; i < (*count) - extra_space; i++)
+			ret[i] = (char *) array[i];
+	}
+
+	return ret;
+}
+
+static char*
+find_environment_var(const char *var)
+{
+	char **s;
+	int len = real_strlen(var);
+
+	len = real_strlen(var);
+
+	for (s = environ; *s; s++)
+		if (real_strncmp(*s, var, len) == 0)
+			return *s;
+
+	return NULL;
+}
+
+static char**
+inject_retrace_env_vars(const char **env)
+{
+	static int follow_children = -1;
+	char **new_env = NULL;
+
+	if (follow_children == -1) {
+		follow_children = 0;
+		if (rtr_get_config_single("forcefollowchildren", ARGUMENT_TYPE_END))
+			follow_children = 1;
+	}
+
+	if (follow_children) {
+		int size;
+#ifdef __APPLE__
+		char *retrace_env_var = "DYLD_INSERT_LIBRARIES";
+#else
+		char *retrace_env_var = "LD_PRELOAD";
+#endif
+		char *var_to_copy;
+
+		new_env = duplicate_string_array((const char **) env, 1, &size);
+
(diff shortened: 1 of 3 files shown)


riataman:
Changing the setting name makes sense.

Not sure about this being the default. Seems like messing with the data to much to make it the default?

Also I just realized that I also need to forward the RETRACE_CONFIG env var if present.

drystone:
That will only be a problem for an exec'd program that sanitises its
environment. If that's the case you'd be unable to trace anyway so use
followexec,0 (if anything, perhaps followexec,0 should be the default so
args are only redirected in non-default case). Personally I think
followexec,1 is the expected behaviour.

followexec,0
followexec,1 (default)

Need similar for fork:

followfork,0
followfork,1 (default)

ronaldtse:
I agree with @drystone , this is an excellent functionality that should be made default. Thanks @riataman  👍

catap:
@drystone but dtrust and strace doesn't follow `exec`/`fork` by default and both of them has special option to change default behaviour.

riataman:
This one is tricky.  We somehow depend on the LD_PRELOAD inheritance rules which as far as I have figured out is always inherit LD_PRELOAD but on the exec functions that allow specifying the environment.

So in pretty much all cases we do follow the children (fork, system, etc) and we can't even take ourselves out of the processes even if we wanted, the only thing we could do is disabling the trace printing, but our code would still be in the middle.

riataman:
I changed the option name to forcefollowexec which is what this does and also made it so that RETRACE_CONFIG env var is also passed in those cases.

I think this is the minimal useful change right now, doing something else and trying to do consistent (if at all possible) would be hard and need a much bigger change that we can leave for another day.

I guess the question is wherer to enable it by default or not. I'm leaning towards not enablebling by default @catap also seems to think the same. While @drystone  and @ronaldtse think should be default.

@erikbor want to break the tie?

erikbor:
If we trace child processes by default then we could run into some problems with system() calls that do stuff such as: system("ps aux | awk '{ print $1 }' | sort -u"); or something like that. Guess that's not quite common but we'd be tracing ps,awk,sort then too which would not be the intended purpose.

I (as a user) would like to specify this myself instead of this being the default behaviour.

(ps: ltrace and gdb also don't follow child processes by default)

ronaldtse:
I'm fine with this not being the default, as long as during the tracing a fork would trigger something (like a colorful message) so I know that there's part of execution I'm not seeing due to the fork, which would prompt me to re-run if necessary.

drystone:
I'm not really concerned with what the default should be, only it should be consistent. Either follow all or not follow all.

@riataman I would have thought 'not following'  would be as simple as removing `RETRACE_*` from the environment before calling `fork` `exec` etc no need to stay in the middle.

exec.c
```diff
@@ -26,8 +26,93 @@
...
+
+
+	/* relying in the fact that calloc initializes memory to zero */
+	ret = (char **) real_calloc(*count, sizeof(char *));
+
+	if (ret) {
+		int i;
+
+		for (i = 0; i < (*count) - extra_space; i++)
+			ret[i] = (char *) array[i];
```
catap: Here we might have small performance optimisation:
1) Use malloc;
2) `ret[i] = NULL` after `for` loop. 

:)

exec.c
```diff
@@ -26,8 +26,93 @@
...
+static char*
+find_environment_var(const char *var)
+{
+	char **s;
+	int len;
+
+	len = real_strlen(var);
+
+	for (s = environ; *s; s++)
+		if (real_strncmp(*s, var, len) == 0)
```
catap: Same here. We're expected that all string is `\0` terminated, so, we can move away from `strlen` and use `strcmp`.

exec.c
```diff
@@ -26,8 +26,93 @@
...
+		char *retrace_config_var = "RETRACE_CONFIG";
+
+		char *var_to_copy;
+
+		new_env = duplicate_string_array((const char **) env, 2, &size);
+
+		var_to_copy = find_environment_var(retrace_env_var);
+
+		if (new_env && var_to_copy)
+			new_env[size - 3] = var_to_copy;
```
catap: I think that I miss something but why `-3`?

exec.c
riataman: I would need a foor loop to set the remaining four spots to zero. In reality calloc is pretty fast due to OS specific tricks (https://stackoverflow.com/questions/2688466/why-mallocmemset-is-slower-than-calloc)

exec.c
riataman: No, the environ global var has the variables in the form of:

"RETRACE_CONFIG=/home/bla/retrace.conf"

And I'm looking to only match on the RETRACE_CONFIG part, thus the strncmp. I could use getenv() but then I'd have to construct the var in the required format rathern than simply point to it.

catap: agree

exec.c
catap: got it

catap:
@riataman I would like to say that enable it by default is fine and better that behaviour of another trace utils.

riataman:
@drystone It doesn't seem to work that way.
```

$ cat test/fork.c
#include <unistd.h>

int main(void)
{
        unsetenv("LD_PRELOAD");
        fork();
        getpid();

        return 0;
}

$ ./retrace ./test/fork
(29686) getenv("OPENSSL_ia32cap", ) = (nil)
(29686) unsetenv("LD_PRELOAD", ) = 0
(29686) fork() = 0
(29687) getpid() = 29687
(29687) fork() = 29687
(29686) getpid() = 29686

(Fixed output a bit by hand since its getting interleaved by the two processes).

```

PR #197, opened 2017-07-05
Disable colors if we aren't logging to a tty

riataman:


riataman:
Thanks @jjr840430 that's a good catch. Will fix it.

PR #201, opened 2017-07-06
fix pledge.c to compile on OpenBSD, remove 'local' from retrace.in because OpenBSD /bin/sh does not like that

erikbor:
. pledge.c had incorrect return_value, and insufficient RETRACE_REPLACE
. retrace.in used local which OpenBSD /bin/sh does not like
. retrace.in used '[ -x ]' to assess executability of '$1' but this does not work when '$1' is not absolute but in PATH


riataman:
Looks good to me. Seems like this went trough a couple of infrastructure changes without being tested in openBSD, which is agree is a pain to do manually.

PR #202, opened 2017-07-06
Crashfix

riataman:
This fixes a crash I ran into


erikbor:
Looks good, tested on Ubuntu and macOS

PR #203, opened 2017-07-06
Time

riataman:
This is not yet finished, but shows the concept. The patch does two things:

1. Print a time stamp on each call. The first call would be 0 and then each subsequent call would show time elapsed since that first call.
2. Print how much time each call took. This can be used as a very crude poor man's profiler. Specially if I put in a config to only show calls that took more than X amount of time.

I still need to make all of this configurable and fix the printing I think the decimal part is wrong as displayed (its actually nano seconds rather than true decimal seconds).


riataman:
image

PR #204, opened 2017-07-06
Fix crashes when running gcc in retrace

riataman:
This was really hard to track down :-\

### DIFF
--- a/common.c
+++ b/common.c
@@ -541,6 +541,7 @@ retrace_event(struct rtr_event_info *event_info)
 		}
 
 		trace_printf(0, ")");
+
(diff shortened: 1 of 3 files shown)


ronaldtse:
Nice stuff @riataman ! Should this (and Firefox) be made into a CI test?

riataman:
Yes I was thinking that running the c compiler under retrace is a nice test.

Not sure about firefox, not that familiar with travis to know how feasible that is. 

I'll work on adding a test for cc

ronaldtse:
Right, and you can compile small / large programs to see if retrace is doing its thing. e.g., the program originally should print out some string but "retracing" gcc makes it print out another string.

PR #206, opened 2017-07-07
Fix OpenBSD build

riataman:
This fixes the build in OpenBSD.

Still retrace won't run when build because it needs to link to pthread. If I manually link it to pthread it works fine. Hopefully @wizzard  can fix the pthread linking.

### DIFF
--- a/common.c
+++ b/common.c
@@ -55,6 +55,7 @@
 #include <arpa/inet.h>
 #include <sys/un.h>
 #include <netdb.h>
+#include <sys/socket.h>
 
 #include "str.h"
 #include "id.h"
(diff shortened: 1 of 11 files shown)


wizzard:
Sure, will take a look today.

wizzard:
Added pthread support.
@riataman with your changes all tests are passing on OpenBSD, good job!

common.c
```diff
@@ -756,7 +754,11 @@ retrace_event(struct rtr_event_info *event_info)
 		return;
 
 	old_trace_state = trace_disable();
+
+	old_trace_state = trace_disable();
```
ronaldtse: Am I reading this right, there are two calls (756, 758) of:
``` c
old_trace_state = trace_disable();
```
?

riataman: Yeah, this one had a bunch of merging issues. I need to go back and rework it.

riataman:
This now works in OpenBSD. Builds in FreeBSD but has a lot of problems, would need to sit down an debug them but I'm not sure that's a priority right now.

PR #213, opened 2017-07-07
Indent calls that where made as a result of another call

riataman:
This is a fun little one. Indent calls that are made as a result of another call (see image).

image

Here you can see that mkfifo calls open and memcpy internally.

### DIFF
--- a/common.c
+++ b/common.c
@@ -117,6 +117,8 @@ static int show_timestamp;
 static int output_file_flush;
 static FILE *output_file;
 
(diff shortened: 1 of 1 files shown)


riataman:
This needs to go back to the oven

ronaldtse:
@riataman in fact I think a JSON output format would be very helpful to allow machine parsing too!

We could separate the output in a formatter pattern so we can have different types of for matters, where the current human readable form is one of them.

PR #221, opened 2017-07-10
Implemented logging options

jjr840430:


common.c
```diff
@@ -1809,3 +1826,159 @@ void *rtr_get_fuzzing_value(enum RTR_FUZZ_TYPE fuzz_type, void *param)
...
+				else
+					opt_val &= !bit_val;
+			}
+		}
+
+		if (opt_type == RTR_LOG_OPT_GRP)
+			g_logging_config.group_bitwise = opt_val;
+		else
+			g_logging_config.stracing_group_bitwise = opt_val;
+	} else if (opt_type == RTR_LOG_OPT_LEVEL) {
```
ronaldtse: @jjr840430 could we short circuit this here like this:

```c
	if (opt_type == RTR_LOG_OPT_GRP)
		g_logging_config.group_bitwise = opt_val;
	else
		g_logging_config.stracing_group_bitwise = opt_val;

	return;
}

if (opt_type == RTR_LOG_OPT_LEVEL) {
```

common.c
```diff
@@ -1809,3 +1826,159 @@ void *rtr_get_fuzzing_value(enum RTR_FUZZ_TYPE fuzz_type, void *param)
...
+	if (stack_trace) {
+		/* check excluded list */
+		if (g_logging_config.stracing_disabled_funcs &&
+			rtr_check_config_token(event_info->function_name, g_logging_config.stracing_disabled_funcs, sep, &reverse) &&
+			!reverse)
+			return 0;
+
+		if ((event_info->function_group & g_logging_config.stracing_group_bitwise) == 0x00)
+			return 0;
+	} else {
```
ronaldtse: @jjr840430 could we short circuit this here like this:

```c
if (stack_trace) {
  ...
  return 1;
}

/* check fuzzing or redirection is applied */
if ((event_info->logging_level & RTR_LOG_LEVEL_FUZZ) || ...)
```

jjr840430: @ronaldtse just fixed. :)

ronaldtse:
Thank you @jjr840430 !

PR #233, opened 2017-07-13
Adds double-free checking

drystone:
So the jury is out as to whether this is worth having, I thought I'd attempt it anyway. This simply sets a bit corresponding to the address returned by `malloc` `realloc` and `calloc` and logs before `free` (or `realloc` if the bit is zero (ie already freed or not allocated). Bitmaps each cover 64k of heap and are held in a singly linked list so maps only exist if `malloc` uses that particular 64k memory region. There is no reporting of unfreed blocks - I suppose an `atexit` handler would need to be installed for this purpose.
Tested with `./retrace gcc -ot t.c` with minimal impact on runtime.

### DIFF
--- a/malloc.c
+++ b/malloc.c
@@ -31,6 +31,70 @@
 #include <sys/mman.h>
 #include <unistd.h>
 #include <errno.h>
+#include <sys/queue.h>
+#include <assert.h>
+#include <pthread.h>
+
+#define BLKSZ 65536
+#define ALIGNSZ (32 / 8)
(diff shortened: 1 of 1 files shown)


ronaldtse:
Very nice @drystone and happy to see you back active 👍

drystone:
Thanks @ronaldtse :) I never was fully away, I've been quietly having a go at [RPC](
https://github.com/drystone/retrace/tree/rpc/rpc)

PR #238, opened 2017-07-14
Added net-funzzing functions

jjr840430:


common.c
```diff
@@ -1987,3 +1961,143 @@ static int rtr_check_logging_config(struct rtr_event_info *event_info, int stack
...
+		init_netfuzz_config();
+
+	/* get index of socket function */
+	for (func_idx = 0; net_func_names[func_idx] != NULL; func_idx++) {
+		if (real_strcmp(func_name, net_func_names[func_idx]) == 0)
+			break;
+	}
+
+	/* check matched status */
+	if (func_idx < 0)
```
drystone: I'm not sure this can ever happen :) You mean `if (net_func_names[func_idx] == NULL)` Also `net_func_names` and `enum RTR_NET_FUNCTION_IDX` could easily get out of sync. Perhaps something like:

```
struct {
    enum RTR_NET_FUNC_ID func_id ;
    const char *name;
} net_funcs = [{
    {NET_FUNC_SOCKET, "socket"},
    {NET_FUNC_CONNECT, "connect"},
    ...
    {NET_FUNC_END, NULL}
};

enum RTR_NET_FUNC_ID func_id = NET_FUNC_END;
int i;
for (i=0; net_funcs[i].func_id != NET_FUNC_END; i++) {
    if (real_strcmp(func_name, net_funcs[i].name) == 0) {
        func_id = net_funcs[i].id;
        break;
    }
}
if (func_id != RTR_NET_FUNC_END)
    return 0;
```

jjr840430: @drystone Great advise 👍

netdb.c
```diff
@@ -80,11 +93,20 @@ struct hostent *RETRACE_IMPLEMENTATION(gethostbyaddr)(const void *addr, socklen_
 
 	retrace_log_and_redirect_before(&event_info);
 
-	hent = real_gethostbyaddr(addr, len, type);
-	if (hent) {
-		event_info.return_value_type = PARAMETER_TYPE_STRUCT_HOSTEN;
-	} else
-		event_info.logging_level |= RTR_LOG_LEVEL_ERR;
+	if (rtr_get_net_fuzzing(event_info.function_name, &err)) {
```
drystone: Why not `rtr_get_net_fuzzing(NET_FUNC_GETHOSTBYADDR, &err)` so `rtr_get_net_fuzzing` wouldn't need to do any of the `strcmp` loop (discussed above). Also could `rtr_get_net_fuzzing` etc. go into new files `netfuzz.c` and `netfuzz.h`?

jjr840430: You are right, @drystone ,
I agreed to make new files.

erikbor:
Hi JJ,

Building on macOS works but the resulting retrace crashes:
```sh
$ ./retrace -f examples/net-fuzzing/netfuzzing.conf examples/net-fuzzing/test_client
test_client(93682,0x7fffabb543c0) malloc: *** malloc was initialized without entropy
./retrace: line 68: 93682 Illegal instruction: 4  RETRACE_CONFIG=examples/net-fuzzing/netfuzzing.conf DYLD_FORCE_FLAT_NAMESPACE=1 DYLD_INSERT_LIBRARIES=.libs/libretrace.dylib examples/net-fuzzing/test_client
```

```sh
$ ./retrace ~/tmp/curl www.google.com
curl(93705,0x7fffabb543c0) malloc: *** malloc was initialized without entropy
./retrace: line 68: 93705 Illegal instruction: 4  RETRACE_CONFIG= DYLD_FORCE_FLAT_NAMESPACE=1 DYLD_INSERT_LIBRARIES=.libs/libretrace.dylib tmp/curl www.google.com
```

```sh
$ ./retrace ~/tmp/ping localhost
ping(93718,0x7fffabb543c0) malloc: *** malloc was initialized without entropy
./retrace: line 68: 93718 Illegal instruction: 4  RETRACE_CONFIG= DYLD_FORCE_FLAT_NAMESPACE=1 DYLD_INSERT_LIBRARIES=.libs/libretrace.dylib tmp/ping localhost
$ 
```

examples/net-fuzzing/README.md
```diff
@@ -0,0 +1,37 @@
...
+NO_MEMORY, LIMIT_SOCKET
+
+2. connect
+ADDR_INUSE, NET_UNREACHABLE, CONN_TIMEOUT
+
+3. bind, listen
+ADDR_INUSE
+
+4. send, sendto, sendmsg
+CONN_RESET, NO_MEMORY
```
catap: you missed that write can also be used to socket

examples/net-fuzzing/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+5. recv, recvfrom, recvmsg
+CONN_REFUSE, NO_MEMORY
```
catap: you missed that write can also be used to socket

examples/net-fuzzing/README.md
catap: does `NO_MEMORY` means that part of buffer was send and send returns something bigger that 0 and less that buffer size that should be send.

netdb.c
jjr840430: @drystone Fixed

examples/net-fuzzing/README.md
jjr840430: @catap read()/write() calls are used for socket communications, but I didn't add fuzzing for these functions.
I think that "incomplete" option is enough to test fuzzing for read/write() calls.

For NO_MEMORY, it's for case when queue is full or kmalloc has failed on kernel space.

riataman:
@jjr840430 I put in a commit that makes it not crash in macos.

jjr840430:
@riataman Great, thanks 👍

netfuzz.c
```diff
@@ -0,0 +1,189 @@
...
+
+/* get network fuzzing error value */
+int rtr_get_net_fuzzing(enum RTR_NET_FUNC_ID func_id, int *err_val)
+{
+	int fuzz_typeset;
+	int matched = 0;
+
+	/* check function ID */
+	if (func_id >= NET_FUNC_ID_MAX)
+		return 0;
```
drystone: This test isn't necessary. You can assume it's called correctly (I'd hope :))

jjr840430:
@drystone No problem, Thanks for your reviews.

jjr840430:
@catap @riataman Can you help me to solve some issues on macos?

It never load configuration and I didn't find why it's not working

riataman:
memoryfuzzing doesn't work in macOS because macOS initialization code isn't tolerant to malloc failing.

examples/net-fuzzing/README.md
catap: @jjr840430 ok, but I think that we can understand that this fd is network or IO related and use `incompleteio` or `net-funzzing`.

But we can keep it as is.

catap:
@riataman yes, it is. But it works but with fixed seed or very small percent of fuzzing :)

PR #239, opened 2017-07-14
Use RTR_CONFIG_START to signal start of the config

riataman:
We were using NULL both to signal the end of the config and the start.

Now we need to initialize RTR_CONFIG_HANDLE to RTR_CONFIG_START


jjr840430:
@riataman Thanks for your help.

PR #244, opened 2017-07-18
Fix for OpenBSD

riataman:
pthread_mutex_lock calls malloc in OpenBSD which gets into a loop
force load the config file before pledge()

### DIFF
--- a/malloc.c
+++ b/malloc.c
@@ -54,7 +54,12 @@ static int map_bit(const void *p, int set)
 	struct map *map, *prev;
 	unsigned long int address;
 	unsigned int offset, bit;
-	int old, mask;
+	int old, mask, old_trace_state;
+
+	if (!get_tracing_enabled())
+		return 0;
+
+	old_trace_state = trace_disable();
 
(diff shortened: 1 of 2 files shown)


erikbor:
Nice one Pablo, it works!

```sh
$ cat ~/pledge.c                                                                                                                                                                                         
#include <unistd.h>
#include <stdio.h>

int
main()
{
        if (pledge("stdio rpath", NULL) == -1)
                err(1, "pledge");

        printf("foobar\n");

        return(0);
}
$ gcc ~/pledge.c -o ~/pledge  
$ ./retrace ~/pledge                                                                                                                                                                                     
(94250) pledge("stdio rpath", ) = 0
(94250) malloc(65536, ) = 0x1f3950486000
foobar
(94250) exit(0, )
$ 
```

riataman:
@erikbor that's actually weird. the printf wasn't logged

riataman:
Interestingly the compiler optimizes a printf call that has a constant string with no formating to puts, which we weren't handling.

PR #247, opened 2017-07-24
Adds beginnings of RPC version

drystone:
I'm finally adding a basic RPC version of retrace in the rpc directory. It's self contained and can be added to the build with `./configure --enable-rpc`. In the rpc directory try something like `./retrace /bin/ls`

Nearly all of the retrace shared library is generated automatically. A rudimentary front end, `retrace.c`, creates a pair of sockets and fork-execs the program under test. It then receives messages over the socket and inspects/prints the call info.

It's a working proof of concept lacking all the niceties of the current retrace (threads, redirection, fuzzing) but all this can fairly easily be ported if there is appetite to do so.


ronaldtse:
Thanks @drystone, this is exciting stuff!

@erikbor could you have a look if anything needs to be changed?

drystone:
Thanks @ronaldtse it's shaping up. I think that by the end of the week it should be handling threads, configuration and some redirections. By which time all underlying problems/oversights should be resolved and we can hopefully port the remaining functionality.

PR #256, opened 2017-08-08
added stringinjector function into retrace

jjr840430:


ronaldtse:
Aha, build failed due to typo. @jjr840430 could you help fix? Thanks.

```sh
fe04237 Added documentation for stringinject config
WARNING: 'charactor' may be misspelled - perhaps 'character'?
#51: FILE: examples/stringinject/README.md:35:
```

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+		real_memcpy(p, buffer, offset);
+
+	real_memcpy(&p[offset], &hex_value, 1);
+	if (offset < len)
+		real_memcpy(p + offset + 1, buffer + offset + 1, len - offset - 1);
+
+	real_memcpy(buffer, p, len);
+
+	/* free buffer */
+	real_free(p);
```
drystone: Could lines 222 - 238 be condensed to `char *p = buffer; p[offset] = hex_value;` or `((char *)buffer)[offset] = hex_value`?

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+
+	int fmt_count, overflow_buf_len;
+
+	/* parse param */
+	p = real_strchr(param, ':');
+	if (!p)
+		return -1;
+
+	memset(param_val, 0, sizeof(param_val));
+	real_strncpy(param_val, param, p - param);
```
drystone: To avoid possible overflow:
```
if (!p || p - param >= 512)
    return -1;

real_strncpy(param_val, param, p - param);
param[p - param] = '\0';
```

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+
+	/* check offset */
+	if (*offset >= len)
+		*offset = len - 1;
+
+	/* inject string by types */
+	switch (type) {
+	case STRINJECT_TYPE_HEX:
+		if (real_strcmp(param_val, "RANDOM") == 0)
+			((char *) val)[0] = rand() % 0xFF;
```
drystone: `rand() & 0xff` or `rand() % 0x100`

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+		else if (real_strncmp(param_val, "0x", 2) == 0) {
+			char hex_val;
+
+			hex_val = strtol(param_val + 2, NULL, 16);
+			if (hex_val == LONG_MIN || hex_val == LONG_MAX)
+				return -1;
```
drystone: Test is a little pointless as hex_val is a char so is unlikely to be LONG_{MIN,MAX}. I'd use
```
unsigned long hex_val;
hex_val = strtoul(...);
if (hex_val > 0xff)
    return -1;
((char *)val)[0] = (char)hex_val;
```

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+		real_memcpy(p + offset + count * 2, buffer + offset + count * 2, len - offset - count * 2);
+
+	real_memcpy(buffer, p, len);
+
+	/* free buffer */
+	real_free(p);
+
+	/* free format string buffer */
+	real_free(fmt_str);
+
```
drystone: 259 - 285 could be condensed to something like:
```
char *p = (char *)buffer + offset;
for (i = 0; i < count; i++) {
    *(p++) = '%';
    *(p++) = 's';
}
```

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+	if (offset < len)
+		real_memcpy(p + offset + count, buffer + offset + count, len - offset - count);
+
+	real_memcpy(buffer, p, len);
+
+	/* free buffer */
+	real_free(p);
+
+	/* free format string buffer */
+	real_free(overflow_buf);
```
drystone: and again with `*(p++) = 'A'`

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+		return 0;
+
+	/* get whole line count of file */
+	while (real_fgets(line_buf, sizeof(line_buf), fp) != NULL)
+		line_count++;
+
+	real_fseek(fp, 0, SEEK_SET);
+
+	/* get line buffer with randomely selected line number */
+	selected_line = rand() % line_count;
```
drystone: division by 0 if line file is emty

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+				break;
+
+			/* remove end line */
+			line_buf[real_strlen(line_buf) - 1] = '\0';
+
+			/* get injection length */
+			inject_len = real_strlen(line_buf) < (len - offset) ? real_strlen(line_buf) : (len - offset);
+
+			break;
+		}
```
drystone: Some problems
- 379 will never be reached (fgets will return NULL if no chars and no \n)
- 382 will lose last char of last line if it isn't `\n` terminated
```
if (line_count == selected_line) {
    inject_len = real_strlen(line_buf);
    if (line_buf[inject_len - 1] == '\n')
        --inject_len;
    if (inject_len > len - offset)
        inject_len = len - offset;
    break;
}
```

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+		real_memcpy(p, buffer, offset);
+
+	real_memcpy(&p[offset], line_buf, inject_len);
+	if (offset < len)
+		real_memcpy(p + offset + inject_len, buffer + offset + inject_len, len - offset - inject_len);
+
+	real_memcpy(buffer, p, len);
+
+	/* free buffer */
+	real_free(p);
```
drystone: No buffers needed, plain `memcpy((char *)buffer + offset, line_buf, inject_len);`

strinject.c
```diff
@@ -0,0 +1,462 @@
...
+
+		if (inject_type == STRINJECT_TYPE_UNKNOWN)
+			continue;
+
+		/* get function list */
+		for (i = 0; i < STRINJECT_FUNC_MAX; i++) {
+			if (rtr_check_config_token(g_inject_funcs[i].name, func_list, "|", &reverse) &&
+				!g_strinject_infos[i].exist) {
+				g_strinject_infos[i].type = inject_type;
+				real_strcpy(g_strinject_infos[i].param, inject_param);
```
drystone: buffer overflow possibility?

jjr840430:
@drystone Thanks for your reviews. :).
It was really good reviews. Can you try review again?

jjr840430:
@drystone You are right. I pushed changes.

drystone:
@jjr840430 I thought you meant to have 0-255 which would be `rand() & 0xff` (not `%`). Though I think 0x7f would be useful, perhaps as another option - `RANDOM` (0-255) or `ASCII` (0-127)  - you'd still want `rand & 0x7f` though :)

jjr840430:
@drystone Can you review this link?
https://github.com/riboseinc/retrace/pull/256/commits/5852c164241bc8d9f208de088c89c993a32e548c

drystone:
@jjr840430 I think you want the full range of hex 0x00 - 0xff. `rand() % 0xff` gives 0x00 - 0xfe. That's what I was initially pointing out. We want the full range to fuzz binary data (`rand() & 0xff` or `rand() % 0x100`). However, `rand() % 0x80` or `rand() & 0x7f` will give ASCII values 0-127 which will be more useful perhaps for text files.

jjr840430:
@drystone so do I need to check whether source buffer is binary or ascii stream before injection?

drystone:
@jjr840430 no I reckon it could be done in the config:

> ## Injection of single hex
> ```
> stringinject,INJECT_SINGLE_HEX,[function name list],[hex value:offset],[injection rate]
> ```
> The hex value may be 0x00 ~ 0xFF or RANDOM *(random 0x00 ~ 0xFF) or ASCII (random 0x00 ~ 0x7F)*.
> The offset may be integer or RANDOM.

Nice piece of documentation BTW :)

strinject.c
ronaldtse: Is this still relevant?

strinject.c
ronaldtse: Is this still relevant?

jjr840430:
@drystone Ah, understand. If it's possible to specify by configuration, then it would be more flexible. 
I will update as you said.

strinject.c
drystone: This is fixed by above `return 0;`

jjr840430:
@drystone @ronaldtse The issues from review were fixed completely. Thanks.

strinject.c
```diff
@@ -0,0 +1,397 @@
...
+	/* check offset */
+	if (*offset >= len)
+		*offset = len - 1;
+
+	/* inject string by types */
+	switch (type) {
+	case STRINJECT_TYPE_HEX:
+		if (real_strcmp(param_val, "RANDOM") == 0)
+			((char *) val)[0] = rand() % 0xff;
+		else if (real_strcmp(param_val, "ASCII") == 0)
```
drystone: `rand() % 0x100`

strinject.c
```diff
@@ -0,0 +1,397 @@
...
+			((char *) val)[0] = rand() % 0x7f;
```
drystone: `rand() % 0x80`

strinject.c
jjr840430: @drystone You are right, That's my fault.
So for ASCII, is it rand() % 0x80, right?

jjr840430:
@drystone fixed.

erikbor:
@jjr840430 

This relates to ```tools/stringinjector/stringinjector```:

## bug 1
After building on macOS I can't get the correct number of inserts.

Here I am expecting 8 inserted bytes:
```
$ echo "BBBB" > test.txt
$ ./stringinject test.txt test-2.txt -o 2 8
$ hexdump -C test-2.txt 
00000000  42 42 41 41 41                                    |BBAAA|
00000005
```
But I only got 3.

Same with the format specifier:
```
erickborsboom@OF-OSX-007:~/src/riboseinc/retrace/tools/stringinjector stringinjector $ ./stringinject test.txt test-2.txt -f 2 2
erickborsboom@OF-OSX-007:~/src/riboseinc/retrace/tools/stringinjector stringinjector $ hexdump -C test-2.txt 
00000000  42 42 25 73 25                                    |BB%s%|
00000005
```

## bug 2
The insert is a complete replace:
```
$ ./stringinject test.txt test-2.txt -o 2 4
$ hexdump -C test-2.txt 
00000000  42 42 41 41 41 41 43 43  44 44 44 44 45 45 45 45  |BBAAAACCDDDDEEEE|
00000010  0a                                                |.|
00000011
```
The last two 'BB's and first two 'CC's have been fully replaced. The expected behaviour would be:
```
$ hexdump -C test-2.txt 
00000000  42 42 42 42 41 41 41 41  43 43 43 43 44 44 44 44  |BBBBAAAACCCCDDDD|
00000010  45 45 45 45 0a                                    |EEEE.|
00000015
```

I can imagine situations where the byte at the start of the offset should also be replaced. Could you add a switch to replace the offset byte when in '-o' and '-f' mode?

ronaldtse:
@jj840430 as mentioned by @erikbor could you add these tests to Travis to prevent future failures? Thanks!

jjr840430:
@erikbor For bug1, stringinjector will not expand the size of original data while replacing.
So if injection length is longer than original, then it will be truncated.

Also 'echo' statement writes endline '\n'. So 3 bytes were replaced from your testing.
Let me know what I'm wrong on implementation of stringinjector, then I will fix soon.
Thanks.

erikbor:
@jjr840430 the offset indeed cannot be larger than the file size of the source, but to test for buffer overflows we should be able to inject long strings without replacing the source content (with the exception of the first byte of the offset).

Imagine an XML file we will stringinject to fuzz an XML parser:
```
<test>
    <header>foo</header>
    <body>bar</body>
<test>
```
When we start overwriting the entire content at say offset 16 then the entire XML file becomes corrupt and some simple sanity checking of the parser will detect that and thus the file is unusable. If we insert a string at offset 16 while leaving the rest of the file in tact then we can test the parser it's XML tag handling without corrupting the entire file.

jjr840430:
@erikbor Understand, so do I need to insert string without replacing when specified -o and -f option?

erikbor:
@jjr840430 yeah that would be great! Could you then also add a switch to choose whether to replace the offset byte?

E.g.:
With replacing the offset byte:
```
$ echo "BBBBCCCCDDDD" > test.txt
$ ./stringinject test.txt test-1.txt -o 3 4 -r
$ hexdump -C test-1.txt 
00000000  42 42 42 41 41 41 41 43  43 43 43 44 44 44 44 0a  |BBBAAAACCCCDDDD.|
00000010
```

Without replacing the offset byte:
```
$ echo "BBBBCCCCDDDD" > test.txt
$ ./stringinject test.txt test-1.txt -o 3 4
$ hexdump -C test-1.txt 
00000000  42 42 42 42 41 41 41 41  43 43 43 43 44 44 44 44  |BBBBAAAACCCCDDDD|
00000010  0a                                                |.|
00000011
```

I want to ensure we don't overwrite 0x00 thus not properly terminating strings, and I there could be situations where we don't want to replace the offset byte (e.g. a space, or ';', or '|' etc).

Cheers!

jjr840430:
@erikbor OK, will update soon.

drystone:
@jjr840430 @erikbor possibly more generally the number of bytes to remove - a bit like JavaScript splice? Something like `-o <offset> <insertlen> [<removelen>]`

jjr840430:
@erikbor Fixed.

jjr840430:
@erikbor do you have any issues with latest build?

erikbor:
@drystone your proposal of replacing number of bytes actually makes a lot of sense, that way the filesize and structure (depending on len) will be kept in tact. I'll add that as a new issue so we can merge this one.

erikbor:
@drystone, I just pulled the latest commit and the replacement functionality is already implemented:
```sh
$ echo "BBBBCCCCDDDD" > test.txt
$ ./stringinject test.txt test-1.txt -o 4 4 -r
$ hexdump -C test-1.txt
00000000  42 42 42 42 41 41 41 41  44 44 44 44 0a           |BBBBAAAADDDD.|
0000000d
```
This really rocks.

@jjr840430 stringinject works like a charm, lets merge this one. For any other issues I'll create issues if they pop up.
Thanks for your work!!!

jjr840430:
Thanks, @erikbor , @drystone .

PR #266, opened 2017-08-13
Bugfix and simplify get_config()

drystone:


jjr840430:
@drystone Thanks for your help. Will test and let you know soon. :+1:

jjr840430:
@drystone Not working... :(
Can you try test on MacOS with stringinjector branch? It works on Ubuntu well, but not on mac.

jjr840430:
@drystone I don't understand well about get_config() function.
Used static variable and not sure why it needs.

drystone:
@jjr840430 I'm not sure how best to test on MacOS without a Mac ... other than Travis. Is there a ci test for the broken functionality?

get_config() has two statics, `pconfig == NULL` until the configuration is loaded at which point it points to the configuration list. It needs to be static to avoid re-reading the config every time. And `config_end` is a static config entry that marks the end of the config list.

jjr840430:
@drystone Here is the link to install macOS VM on your Windows system.
https://techsviewer.com/install-macos-sierra-vmware-windows/

Issue #237, opened 2017-07-14
rtr_get_config_multiple() issue


jjr840430:

jjr840430:
infinite loop when using rtr_get_config_multiple()

The example code:
while (1) {
 r = rtr_get_config_multiple(...);
 if (r == 0)
   break;
 else
   do_something;                 /* not break from while statement */
}

riataman:
this happens when where's no option after the one you are getting. as a quick work around you can add another option at the end of the file.

riataman:
Fixed by https://github.com/riboseinc/retrace/pull/239

now initialize it like this:

RTR_CONFIG_HANDLE config = RTR_CONFIG_START;

drystone:
I added this test to test/config.c and it fails
```
        ch = RTR_CONFIG_START;
        i = 0;
        do { 
                r = rtr_get_config_multiple(&ch, "config-test2", ARGUMENT_TYPE_STRING, ARGUMENT_TYPE_INT,
                    ARGUMENT_TYPE_END, &s, &i);
                ++i;
                assert(i < 10);
        } while (r);
```

jjr840430:
@drystone I think the testing code should be as the followings:
```
        ch = RTR_CONFIG_START;
        i = 0;
        do {
        	int j;

                r = rtr_get_config_multiple(&ch, "config-test2", ARGUMENT_TYPE_STRING,  ARGUMENT_TYPE_INT,
                    ARGUMENT_TYPE_END, &s, &j);

                ++i;
                assert(i < 10);
        } while (r && ch != NULL);

```

drystone:
@jjr840430 that's a good spot :) But it still breaks - I'll look at your latest PR

jjr840430:
@drystone Still not working. I think the testing code for config isn't enough to test that function.

drystone:
@jjr840430 so what's happening - is this Mac specific or more general?

jjr840430:
@drystone It's Mac specific issue.

PR #268, opened 2017-08-15
fixed config parser

jjr840430:


jjr840430:
@drystone I changed get_config() function. It seems have some issue while using STAILQ functions.
I converted to use linked list for config parsing.

jjr840430:
@erikbor Can you test stringinjector function in this branch? It would be worked on MacOS.
Now I'm checking issues from Travis CI checker.

erikbor:
@jjr840430 I did a quick test and it works beautifully:
```
(24546) sendto(0 [connected[172.217.24.36:80]], 0x7fff5f931300, 12, 0xc00000000, 0, ) = 12
	0000000	 7465 7374 4141 4141 7465 7374                     | testAAAAtest
(24546) send(0 [connected[172.217.24.36:80]], 0x7fff5f931300, 12, ) = 12 [[redirected]] [fuzzing seed: 1502790944]
	0000000	 7465 7374 4141 4141 7465 7374                     | testAAAAtest
(24546) memcpy(0x7fff4e000a00, 0x1002cefa6, 5, ) = 0x7fff4e000a00
sent
```

I can't test any further right but I will do so tomorrow.

Many thanks for fixing this!!!

jjr840430:
Thanks, @erikbor

drystone:
@jjr840430 It would be preferable to find the config bug rather than to rewrite get_config
- STAILQ is recommended rather than roll-your-own linked lists. <queue.h> rocks
- config lines are now limited length where previously they could be any length
- I suspect [#237](https://github.com/riboseinc/retrace/issues/237) or something like is reintroduced (the test/config.c `(r != 0)` was correct as far as I'm aware, the 4th `config-test2` read should return 0 (there are only 3 in the file)
Is there any chance you can isolate the bug and fix it. I suspect the config parser was already working and the OS X problem is related to some other heap/stack corruption.

jjr840430:
@drystone 

The configuration for 'config-test2' have only 3 lines, so 4-th calling of rtr_get_config_multiple will return 0.
So I think rtr_get_config_multiple() function didn't work properly so far.

Also, Mac OS didn't call runtests.sh when launching Travis CI build. i.e. it didn't test config parser unit test so far.
Here is the link to show it.
https://travis-ci.org/riboseinc/retrace/jobs/264688734

STAILQ is good solution to manipulate the queue list.
However, I think we don't need to use it, as long as we won't use several complex queue operations.
Also the functions for STAILQ may contain several syscall APIs such as malloc, memcpy, free.
So it's not prefer way than using linked list.

I think, if the size of configuration line is 512, then it's enough.
There is no use to have a very long configuration.

If my thoughts are wrong, then let me know.

drystone:
@jjr840430 thanks for your consideration.

Yes, you a quite right, the 4th call to rtr_get_config_multiple() should (and did) return 0. Thats why the test originally failed when non-zero (`FAIL(r != 0`). Now it is broken and loops through the list indefinitely. You've fixed this by adding `if (!config) break;` throughout the codebase. So now we've a test to break on return from get_config_multiple and another before calling it again ie. the behaviour of get_config_multiple is markedly different and we now need to remember to add the additional test after processing each config line (that's code smell IMHO).

Nice one for spotting the MacOS tests weren't working. I hope the Mac folks will be able to remedy this.

I think you may be wrong about queue.h - it's simply macros to manipulate pointers and its use is recommended in [our coding guidelines.](https://www.freebsd.org/cgi/man.cgi?query=style&sektion=9) (and Mac is BSD based so really should have it right!)

And regarding the line length - yes 512 bytes seems ample but why impose an arbitrary limit when there's really no need?

Do you think the MacOS bug that this is addressing is really fixed or do you think it's still lurking somewhere ... I suspect we'll never know because we don't know what caused it :)

jjr840430:
@erikbor @drystone OK, I will try fix as you mentioned. Thanks.

PR #284, opened 2017-08-25
Fixed strinject issue on mac

jjr840430:


jjr840430:
@erikbor It would be work on Mac without any issues. :)
@drystone Thanks for your help to solve the problems.

examples/stringinject/strinject_test.c
```diff
@@ -60,7 +60,7 @@ static void test_fread(void)
 		if (!fp)
```
drystone: line 57: `size_t len;` to avoid confusion.

src/file.c
```diff
@@ -531,18 +536,31 @@ size_t RETRACE_IMPLEMENTATION(fwrite)(const void *ptr, size_t size, size_t nmemb
 	event_info.return_value_type = PARAMETER_TYPE_INT;
 	event_info.return_value = &r;
 	event_info.logging_level = RTR_LOG_LEVEL_NOR;
-	retrace_log_and_redirect_before(&event_info);
 
-	if (rtr_str_inject(STRINJECT_FUNC_FWRITE, (void *) ptr, size * nmemb)) {
+	if (size == 1 && rtr_str_inject(STRINJECT_FUNC_FWRITE, (void *) ptr, size * nmemb, &inject_buffer, &inject_len)) {
```
drystone: `size == 1` so `size * nmemb == nmemb`

jjr840430: yes, right

drystone: also `fwrite("hello", 5, 1, f)` is probably the same as `fwrite("hello", 1, 5, f)` but only one will get injected. Is that intentional?

jjr840430: yes, not possible to inject using fwrite when size is more than 1
but other functions are possible

src/file.c
```diff
@@ -531,18 +536,31 @@ size_t RETRACE_IMPLEMENTATION(fwrite)(const void *ptr, size_t size, size_t nmemb
...
-	r = real_fwrite(ptr, size, nmemb, stream);
-	if (r < size * nmemb)
+	retrace_log_and_redirect_before(&event_info);
+
+	r = real_fwrite(enable_inject ? inject_buffer : ptr,
+			size,
+			enable_inject ? inject_len : nmemb,
+			stream);
+
+	if (r < (inject_len > 0 ? size * nmemb : size * inject_len))
```
drystone: This looks odd to me - equates to `if (r < 0)` when `inject_len == 0` ?

jjr840430: inject_len is always more than 0

jjr840430: ah, maybe it's wrong, thanks for your finding issue

jjr840430: should be inject_len > size * nmemb

jjr840430:
@drystone ah, I forgot to replace when the type is SINGLE_HEX. I will modify.
But it needs to insert strings for other types.

drystone:
@jjr840430 I don't think it's sensible for a write to write more data than intended - you're messing with the universe :)

jjr840430:
@drystone I don't want to mess with the universe, but it's @erikbor 's request. :)

ronaldtse:
Let's get @erikbor to confirm if he wants to mess with the universe 😉

erikbor:
@jjr840430 works really well! I'm going to fuzz loads of software this week :) Thanks for this really good implementation!!!

erikbor:
@jjr840430 yes we do want to mess with the universe! If we send longer strings than intended using a retraced client then we could potentially cause a buffer overflow in a server/daemon :D

erikbor:
@jjr840430 I also cannot get the INJECT_SINGLE_HEX to work:

```sh
$ cat ~/tmp/send.conf
stringinject,INJECT_SINGLE_HEX,send,3:RANDOM,1
$ ./retrace -f ~/tmp/send.conf ~/tmp/send
<SNIP>
(2837) memcpy(0x7fff4e0008a0, 0x7fff4bc00097, 7, ) = 0x7fff4e0008a0
(2837) sendto(0 [socket[172.217.24.36:80]], 0x7fff5c7dc300, 12, 0, 0x0, 0, ) = 12
	0000000	 7465 7374 7465 7374 7465 7374                     | testtesttest
(2837) send(0 [socket[172.217.24.36:80]], 0x7fff5c7dc300, 12, ) = 12
	0000000	 7465 7374 7465 7374 7465 7374                     | testtesttest
(2837) memcpy(0x7fff4d001e00, 0x103423fa6, 5, ) = 0x7fff4d001e00
sent
<SNIP>
$ ./retrace -f ~/tmp/send.conf ~/tmp/send
(2893) sendto(0 [socket[172.217.24.36:80]], 0x7fff5e7ef300, 12, 0, 0x0, 0, ) = 12
	0000000	 7465 7374 7465 7374 7465 7374                     | testtesttest
(2893) send(0 [socket[172.217.24.36:80]], 0x7fff5e7ef300, 12, ) = 12
	0000000	 7465 7374 7465 7374 7465 7374                     | testtesttest
(2893) memcpy(0x7fff4d800a00, 0x101410fa6, 5, ) = 0x7fff4d800a00
sent
(2893) free(0x0, )
<SNIP>
```

jjr840430:
@erikbor Can't you get working of stringinject?
Please refer the following link for configuration settings.

https://github.com/riboseinc/retrace/tree/master/examples/stringinject

erikbor:
@jjr840430 I got it working now, it was indeed some configuration syntax problem. So far it works really well. The only thing that is missing is the "mess with the universe" buffer overflow sending. Once that's in then we can start fuzzing a lot of daemons :) Cheers

ronaldtse:
Since it's working I'm going to first merge this. @erikbor can you create a new issue on messing with the universe? Thanks!

PR #289, opened 2017-09-01
Implement separating the output in a different file per thread/proces

riataman:


README.md
```diff
@@ -320,12 +320,24 @@ Will redirect a `fopen()` call from `/etc/passwd` to `/tmp/passwd`.
...
 
 ``` sh
-logtofile,retrace.log
+logtofile,retrace.log,1
 ```
 
 Will send the log file to a file rather than `stderr`. You can configure
 log output to write to `/dev/null` disable logging completely.
 
+The second parameter indicates where we should flush the output after every
```
drystone: s/where/whether/

src/common.c
```diff
@@ -195,6 +202,42 @@ retrace_get_time(void)
 	return ret;
 }
 
+static void
+retrace_init_file_key(void)
+{
+	pthread_key_create(&per_thread_logging, NULL);
```
drystone: `pthread_key_create(&per_thread_logging, fclose);`?

erikbor:
@riataman looks good to me

src/common.c
riataman: Unfortunately we would have to live with the memleak. Closing the log file by the pthread_key_create destructor closes it too early while there are still operations that we want to log in that thread going on.

drystone: What `FILE *` are you going to use after pthreads has applied the destructor (if any)? Surely further calls to `pthread_getspecific()` will be undefined at best.

riataman: Well, unfortunately for what we are doing we can't go by defined behavior but by implementation specific holes. In Linux pthread_getspecific seems to work EVEN after the destructor has been called. This is how I noticed we were still logging stuff after I fclosed the file in the descriptor. Because I started getting all kind of warnings about using unused memory in valgrind.

Issue #163, opened 2017-06-30
Add warning/notice on system() calls that use no absolute paths


erikbor:
E.g.: system("id"); should show a warning that this is user controllable and thus exploitable in case when the system() call is made by a setuid executable.

E.g.: system("/bin/id"); should not show this warning.

Please take piping and several commands into account. E.g.: system("/bin/ps aux | grep foo"); and system("/bin/id; uname");

riataman:
This is actually going to be very hard to implement we would basically need to implement all forms of escaping done by whatever shell the user has.

riataman:
Taking myself out if somehow wants to give it a shoot :)

PR #290, opened 2017-09-02
Check for possible relative paths in system()

riataman:
Implements this: https://github.com/riboseinc/retrace/issues/163

I figured that even when it may not catch 100% of the cases (since system() basically dumps the command into the shell and that has a lot of escaping, expansions and such) we might as well implement the most common ones and word the warning as "MAY contain a relative path".

### DIFF
--- a/src/exec.c
+++ b/src/exec.c
@@ -28,6 +28,7 @@
 #include "malloc.h"
 #include "str.h"
 #include <unistd.h>
+#include <ctype.h>
 
 
 /* Duplicates a string array adding extra spaces,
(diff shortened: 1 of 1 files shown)


ronaldtse:
@erikbor your comments please 👍

src/exec.c
```diff
@@ -112,6 +113,59 @@ inject_retrace_env_vars(const char **env)
...
+				}
+
+				s = real_strchr(s + 1, '|');
+			}
+		}
+	}
+
+	return retval;
+}
+
```
drystone:
```
static int
may_contain_relative_paths(const char *s)
{
    while (s != NULL) {
        while (*s == ';' || *s == '|' || isspace(*s))
            ++s;
        if (*s && *s != '/')
            return 1; /* could check against list of builtins here ? */
        s = real_strpbrk(s, ";|");
    }
    return 0;
}
```

riataman:
I would take a guess and think that the majority of the cases are a single command and that's it. So the simplest check is checking if the command starts with a / or not.

The next more common case I guess are multiple commands (check a / after a semicolon) and pipes (check for / after |)

Again taking a wild guess I'm thinking that should covert the vast majority  ofl the cases you might see in the wild AND the code seems very simple to me.

Attempting to cover 100% of the cases would be pretty much impossible, since it depends on what shell the user is running and that can be anything.

erikbor:
@riataman nice one Pablo!

PR #291, opened 2017-09-02
Dont print a trailing comma

riataman:
We have been printing a trailing comma like this:

```
(7529) pipe(3, 4, ) = 0
                ^
```

Eliminate that.


ronaldtse:
LGTM, thanks @riataman !

PR #294, opened 2017-09-04
Added injection for SSL read/write

jjr840430:


src/common.c
```diff
@@ -1822,7 +1823,7 @@ static void parse_logging_options(int opt_type, char *opt_str)
 
 		for (i = 0; rtr_logging_groups[i] != NULL; i++) {
 			if (rtr_check_config_token(rtr_logging_groups[i], opt_str, sep, &reverse)) {
-				int bit_val = (i == 0) ? 0x01 : i * 2;
+				int bit_val = exp2(i);
```
drystone: I'm pretty sure `1 << i` does what you want here without needing libm and floating point.

src/file.c
```diff
@@ -531,18 +536,31 @@ size_t RETRACE_IMPLEMENTATION(fwrite)(const void *ptr, size_t size, size_t nmemb
...
-	r = real_fwrite(ptr, size, nmemb, stream);
-	if (r < size * nmemb)
+	retrace_log_and_redirect_before(&event_info);
+
+	r = real_fwrite(enable_inject ? inject_buffer : ptr,
+			size,
+			enable_inject ? inject_len : nmemb,
+			stream);
+
+	if (r < (inject_len > size * nmemb ? size * nmemb : size * inject_len))
```
drystone: Isn't `inject_len` undefined when `size != 1`? Also, `fwrite returns number of items written.
```
if (r < (enable_inject ? inject_len : nmenb))
```

src/sock.c
```diff
@@ -485,18 +489,31 @@ ssize_t RETRACE_IMPLEMENTATION(send)(int sockfd, const void *buf, size_t len, in
...
+	}
+
+	retrace_log_and_redirect_before(&event_info);
 
+	if (!err) {
+		ret = real_send(sockfd,
+			enable_inject ? inject_buffer : buf,
+			enable_inject ? inject_len : len,
+			flags);
 		if (errno)
```
drystone: I've noticed a few places where `if (errno)` is used regardless of the result of the previous system call. My understanding is that the value of `errno` is undefined unless the previous call indicates a failure. In this case `if (ret == -1)` would be a better choice as we're not actually bothered about the value of `errno`

erikbor:
@drystone check out [http-server-overflow](https://github.com/riboseinc/retrace/tree/master/examples/http-server-overflow) for a PoC. I hope that convinces you that with this functionality we will be able to find vulnerabilities in network daemons.

jjr840430:
@drystone Thanks for your reviews. Please check again.

drystone:
@erikbor - I've been watching the webserver coredumps with interest :)

src/common.c
```diff
@@ -60,6 +60,7 @@
 #include <sys/un.h>
 #include <netdb.h>
 #include <sys/socket.h>
+#include <math.h>
```
drystone: I think this is no longer required.

PR #299, opened 2017-09-27
Changes to RPC version

drystone:
Adds --help and --version
Adds stream, dir and fd tracing
Adds socket, open, openat, creat, sendto, recv and recvfrom
Change from single handlers to handler lists
Converts backtrace into postcall handler
Rework log handler overrides
Adds logging for errno, msghdr, sockaddr and iovec
Use real_sendto instead of real_send
FreeBSD compatibility
Fix thread termination loop in OpenBSD

### DIFF
--- a/ci/before_install.sh
+++ b/ci/before_install.sh
@@ -11,16 +11,21 @@ get_os() {
 }
 
 macos_install() {
-	brew upgrade
-	brew install openssl
-	brew install make
-	brew install cmake
-	brew install autoconf
-	brew install automake
-	brew install libtool
-	brew install cmocka
-	brew install pkg-config
-	brew install ruby
+	brew update
+	packages="
+		openssl
+		make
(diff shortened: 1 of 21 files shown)


drystone:
@ronaldtse have you any idea why the Travis macOS is failing? I've tried both `brew update` and `brew upgrade` in the preinstall script and both break before the build - perhaps the base image is outdated ... can we do anything about that or would it be a Travis internal problem?

ronaldtse:
I actually have no idea how Travis builds their macOS builds. As long as the Jenkins one is working, the build should be working...

Perhaps you can check out https://github.com/riboseinc/rnp 's travis configuration? I remember there's a `brew update` line there.

drystone:
Excellent, thanks :) Last night's build was el capitan, this morning's leopard so may still be broken - merging anyway ...

Issue #304, opened 2017-10-04
Strange behaviour of the output of sendto() injection


erikbor:
@drystone can you take a look at this please?

I have an example UDP echo server:

```C
#include <unistd.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <arpa/inet.h>
#include <sys/socket.h>

#define BUFLEN 512
#define PORT 8888

void die(char *s) {
	perror(s);
	exit(1);
}

int main(void)
{
	struct sockaddr_in si_me, si_other;

	int s, i, slen = sizeof(si_other), recv_len;
	char buf[BUFLEN];

	if ((s = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP)) == -1)
		die("socket");

	memset((char *)&si_me, 0, sizeof(si_me));

	si_me.sin_family = AF_INET;
	si_me.sin_port = htons(PORT);
	si_me.sin_addr.s_addr = htonl(INADDR_ANY);

	if (bind(s, (struct sockaddr *)&si_me, sizeof(si_me)) == -1)
		die("bind");

	while (1) {
		printf("Waiting for data...");
		fflush(stdout);

		if ((recv_len = recvfrom(s, buf, BUFLEN, 0, (struct sockaddr *)&si_other, &slen)) == -1)
			die("recvfrom()");

		printf("Received packet from %s:%d\n", inet_ntoa(si_other.sin_addr), ntohs(si_other.sin_port));
		printf("Data: %s\n", buf);

		if (sendto (s, buf, recv_len, 0, (struct sockaddr *)&si_other, slen) == -1)
			die("sendto()");
	}

	close(s);
	return(0);
}
```

Running the echo server in **terminal 1**:
```console
$ ./retrace ../echo
(14222) socket(2, 2, 17) = 3
(14222) vsnprintf("bind[0.0.0.0:8888]", 128, "bind[%s:%d]" -> "bind[0.0.0.0:8888]") = 18
(14222) bind(3 [bind[0.0.0.0:8888]], 0x7ffd3a0942a0, 16) = 0
(14222) malloc(1024) = 0x560184bf0080
(14222) printf("Waiting for data..." -> "Waiting for data...") = 19
Waiting for data...Received packet from 192.168.1.155:40437
```

Sending a string to the echo server in **terminal 2**:
```
$ echo AAAAAAAAAAAAAAAAA | nc -u 192.168.1.155 8888
AAAAAAAAAAAAAAAAA
```

[...]

PR #305, opened 2017-10-04
Bugfix issue #304, send/sendto output

drystone:


erikbor:
@drystone great fix, ouput is good now. Thanks!

Issue #306, opened 2017-10-04
Introduce an overwrite option for stringinject INJECT_FORMAT_STR, stringinject INJECT_SINGLE_HEX and stringinject INJECT_FILE_LINE


erikbor:
The `stringinject` function currently **inserts** `INJECT_FORMAT_STR`, `INJECT_SINGLE_HEX` and `INJECT_FILE_LINE`. So when `buf` is something like `this is my string` (len 17 bytes) and we `stringinject` `INJECT_FORMAT_STR` at offset 2 with a count of 2 the resulting `buf` will be: `thi%s%ss is my string` (len 21 bytes).

Can you please introduce an **overwrite** option to these 3 string injects.
This is to prevent packet sizes from growing and we will be able to surgically replace a particular string in `send()` and `sendto()` etc. The `buf` example mentioned earlier would look something like this: `thi%s%s my string` (len 17 bytes).

@drystone do you have time for this?

erikbor:
In case the `buf` size is not big enough to hold the injected strings then truncating the injected string is alright.

drystone:
@erikbor sure. We'll need to add a `replacecount` parameter (or `replace` boolean). Should I adjust the existing syntax or add a new injection type?

erikbor:
@drystone what ever you think makes the most sense! Thanks a lot, this will be a killer feature :D

erikbor:
@drystone could you also come up with an example for the `/examples` directory? Cheers

drystone:
Yes, will do, currently working on some tests too.

erikbor:
@drystone Hi John, I just had the opportunity to thoroughly test your recent commits and I am wondering how I can define wether the overwrite at pos should happen or not.

Another question I have is what does `STRINJECT_TYPE_STANDARD` do?

Cheers

erikbor:
@drystone John, can you give an update regarding the overwrite option please? We're already using Retrace in real life fuzzing of NSD and it would be great if we can also include fuzzing without altering the packet size. Thanks!

drystone:
@erikbor not got around to documenting yet but meantime, please see comment at #307 and look at [`test/strinject.sh`](https://github.com/riboseinc/retrace/blob/master/test/strinject.sh) and [`test/strinject.c`](https://github.com/riboseinc/retrace/blob/master/test/strinject.c) for reference

erikbor:
[...]

PR #307, opened 2017-10-10
Extend strinject (issue #306)

drystone:
Fixes retrace.in to handle spaces in executable args.
Fixes strinject buffers and log params.
Adds STRINJECT_TYPE_STANDARD.
Adds some strinject testing.


drystone:
@erikbor can you give this a try - I've not documented yet but config lines should be something like
```
stringinject,function,type,control,probability
```
where `function` is `read`, `write` etc, `type` is one of `ascii()`, `random()`, `or(<str>)`, `xor(<str>)`, `and(<str>)` or `chr(<str>)`. `control` is `offset[:[addlen][:[cutlen]]]` so `chr(123),0` will replace the first 3 bytes with `123`, `chr(123),0:7` will replace the first 6 bytes with `1231231`, `chr(123),1,6,0` will insert `123123` at offset 1. Hex values can be supplied using `\nn`. `or`, `and` and `xor` perform logic operations to twiddle bits so `xor(\01),0:6` will toggle the low bit of the first 6 bytes, `or(\01),0:6` will set the low bits and `and(\fe),0:6` will clear the low bits of the first 6 bytes.

ronaldtse:
Thanks @drystone ! @erikbor is this good to go? Thanks!

ronaldtse:
@erikbor says we can merge it. Thanks @drystone !

PR #318, opened 2017-10-29
Fixing issues related with socket system call

jjr840430:


erikbor:
@jjr840430 

Looks good! I tested it with the `http-server-overflow` and that still worked.

src/common.c
```diff
@@ -609,30 +609,42 @@ retrace_print_parameter(struct rtr_print_buf *print_buffer, unsigned int event_t
...
 	case PARAMETER_TYPE_STRUCT_SOCKADDR:
-		switch ((*(struct sockaddr **) *value)->sa_family) {
+	{
+		struct sockaddr *addr = *((struct sockaddr **) *value);
+
+		char ipaddr[INET6_ADDRSTRLEN];
+		int port;
+
+		if (addr == NULL)
+			break;
```
drystone: Can we `trace_printf(print_buffer, "(nil)")` here (as for `NULL` string)?

src/common.c
```diff
@@ -733,6 +745,168 @@ retrace_print_parameter(struct rtr_print_buf *print_buffer, unsigned int event_t
...
+
+			{-1, NULL}
+		};
+
+		for (i = 0; socket_opts[i].opt_str != NULL; i++) {
+			if (socket_opts[i].opt == opt)
+				opt_str = socket_opts[i].opt_str;
+		}
+
+		trace_printf(print_buffer, 0, "%s(%d)", opt_str, opt);
```
drystone: I think a test for non-null `opt_str` would be sensible otherwise this code relies on undefined `printf` behaviour for NULL pointer

src/common.c
```diff
@@ -733,6 +745,168 @@ retrace_print_parameter(struct rtr_print_buf *print_buffer, unsigned int event_t
...
+			{SOL_SOCKET, "SOL_SOCKET"},
+			{-1, NULL}
+		};
+
+		for (i = 0; proto_levels[i].level_str != NULL; i++) {
+			if (proto_levels[i].level == level)
+				level_str = proto_levels[i].level_str;
+		}
+
+		trace_printf(print_buffer, 0, "%s(%d)", level_str, level);
```
drystone: Same as for `opt_str` above

src/common.h
```diff
@@ -45,59 +45,66 @@ enum RTR_FUZZ_TYPE {
...
+	PARAMETER_TYPE_STRUCT_SOCKADDR,		/* struct sockaddr */
+	PARAMETER_TYPE_FD_SET,				/* fd_set: char **set, int *nfds, fd_set **in, fd_set **out */
+	PARAMETER_TYPE_STRUCT_HOSTEN,		/* struct hosten */
+	PARAMETER_TYPE_IP_ADDR,				/* ip addr: void **addr, int *type */
+	PARAMETER_TYPE_STRUCT_ADDRINFO,		/* struct addrinfo */
+	PARAMETER_TYPE_SOCKET_DOMAIN,		/* socket domain */
+	PARAMETER_TYPE_SOCKET_TYPE,			/* socket types */
+	PARAMETER_TYPE_SOCKET_OPTION,		/* socket option */
+	PARAMETER_TYPE_PROTO_LEVEL
+};
```
drystone: Personally I preferred the `#defines` in this case because of the bitwise operations with `PARAMETER_FLAG_`s and confusing `enum` values that aren't explicitly listed in the declaration.

jjr840430:
@drystone Thanks for your reviews. I will look and let you know :)

src/common.c
jjr840430: @drystone `opt_str` was initialized at declaration as `SOCKET_OPTION`, but I will try to add the checking statement whether `opt` has `-1`

src/common.h
jjr840430: @drystone I didn't checkout that case, will revert to `#define`

jjr840430:
@drystone All issues were fixed. Please check them, Thanks.

drystone:
Thanks @jjr840430 - all good. I'll merge once Travis is happy and if you haven't already done so :)

jjr840430:
@drystone Thanks, Travis is OK for now :)

PR #324, opened 2017-11-02
Added readv/writev tracing

jjr840430:


jjr840430:
@drystone That makes sense. I think your option is better.

jjr840430:
@drystone Thanks for your reviews. OK, I will merge it soon. Thanks.

PR #327, opened 2017-11-07
fixed some issue related with fmt and overflow fuzzing

jjr840430:


erikbor:
@jjr840430 looks good, thanks for adding `axfr` to the `.gitignore` that was my fault.

PR #330, opened 2017-11-12
Added cli extensions for netfuzz module

ikolomi:
1. netfuzz cli commands for statistics and config retrieval
2. definitions for constructor functions priorities: retrace_main.h
3. definitions for environment vars: retrace_env.h

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -270,8 +270,8 @@ Press CTRL-A Z for help on special keys
 Failed to get command id, retry
 Retrace command menu
 \--------------------
-[0] Say Hi
-[1] Terminate process
+[0] <Main> Say Hi
+[1] <Main> Terminate process
 Enter command id>> 
 ----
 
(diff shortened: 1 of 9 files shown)


jjr840430:
@ikolomi I think it would be better to use `real_func` functions rather than using `syscalls` directly.
 For example, you can modify as the following:
```
#define retrace_netfuzz_info(fmt, ...) printf("RETRACE-NETFUZZ [INFO]: " fmt, ## __VA_ARGS__)
```
```
#define retrace_netfuzz_info(fmt, ...) real_printf("RETRACE-NETFUZZ [INFO]: " fmt, ## __VA_ARGS__)
```
@drystone What do you think about it?

jjr840430:
@ikolomi To solve coding style issues, you should merge commits into single.
In `retrace` project, `Travis CI` checks all commits after pushed pull requests.

drystone:
Please look at the ./rpc version of retrace as it solves a lot of these problems by design:

- Client/server model is already in place.
- Minimal impact on traced program as most processing is performed in separate address space.
- Initialisation/configuration is done client side before the traced program is started.
- Most handling of the interposed function calls is done client side so there is no worry about real_ variants.

Issue #334, opened 2017-11-16
loading of the Retrace configuration via an environment variable


erikbor:
Hi @jjr840430 can you implement loading of the Retrace configuration via an environment variable?

Now we can specify a configuration file that can be defined by the environment variable `RETRACE_CONFIG`. Retrace will then open that file and get its configuration.

To make mass fuzzing faster it would be beneficial if we can have the entire configuration in a special environment variable.

This is an example of how we could put the configuration in the environment:
```C
#include <stdio.h>
#include <stdlib.h>

int
main()
{
    char retraceconf[512];

    snprintf(retraceconf, sizeof(retraceconf), "logging-global,LOG_GROUP_NET,LOG_LEVEL_ALL\nstringinject,sendto,chr(\\ff),10:8,1");
    setenv("RETRACE_CONFIG", retraceconf, 1);
    printf("%s\n", getenv("RETRACE_CONFIG"));
    return(0);
}
```

jjr840430:
@erikbor OK, I will implement it. Thanks.

jjr840430:
@erikbor `RETRACE_CONFIG` is already used as an environment variable for specifying the configuration file path.

What about using `RETRACE_CONFIG_VAR` instead it.

erikbor:
Any variable name is fine, so `RETRACE_CONFIG_VAR` is alright. Thanks!

jjr840430:
By the way, is it possible to modify configuration var while tracing?
I think it's possible case when tracing `daemon` processes, so we may update configurations while running `daemon`.

erikbor:
Yeah that will be one of the key modifications that needs to be done soon. I described the first steps in the discussion with you and Ilja. I am writing a more detailed explanation with a diagram and will soon share that for your review.

jjr840430:
Another question, when `retrace` has the configuration file using `-f` option and `RETRACE_CONFIG_VAR` env, then which configuration should be loaded?

erikbor:
Maybe check whether the first line of `RETRACE_CONFIG` results in a successful `stat()` which would mean that's a valid file and then fopen it. If the `stat()` fails then proceed with loading the config from that environment variable?
Do you think that's a good approach?

jjr840430:
I think `RETRACE_CONFIG_VAR` may remain after completed tracing for the specific application.
If another application should be traced, then old `RETRACE_CONFIG_VAR` may bring some conflicts.

[...]

PR #335, opened 2017-11-16
Load configuration via environment variable 'RETRACE_CONFIG_VAR'

jjr840430:
Implemented issue #334

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1575,19 +1575,16 @@ get_config_file()
 	FILE *config_file = NULL;
 	char *file_path;
 	int olderrno;
-	char *config_env;
 
 	char *file_path_user;
 	char *file_name_user = ".retrace.conf";
-
-	int config_fd;
 
 	/* save latest errno */
 	olderrno = errno;
 
 	/* If we have a RETRACE_CONFIG env var, try to open the config file from there. */
 	file_path = real_getenv(RETRACE_ENV_CONFIG_FILE);
-	if (file_path)
+	if (file_path) {
 		config_file = real_fopen(file_path, "r");
 		if (config_file)
 			return config_file;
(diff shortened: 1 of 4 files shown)


jjr840430:
@drystone I think it would be better to modify the configuration of `retrace` at run-time.
Then is it possible to use the wrapper script in this case?

drystone:
@jjr840430 , no it won't be possible to modify config from the script.

I think it would be better to avoid the temporary file creation - you either have to create a temporary file with no way of deleting or use a predefined path (~/.something) that can cause problems if you have more than one instance of `retrace`.

How about modifying `get_config` to add the config from the environment to the config list and then load the config file as before, adding to the list?

It's moot anyway as config is about to receive a major overhaul.

ikolomi:
Guys,
Agree with @drystone . To summarize:
- IMHO, config via env should not replace the config via file, but complement
- Why to create the temporary file in the first place?  The config setup can be as following:

1. look for the configuration env. variable. if exists parse and setup config, done
2. look for the file configuration env. variable. if exists parse the file and setup config, done
3. Use the default config, done

jjr840430:
Agreed, guys.
I didn't regard the case of multiple instances of retrace. Thanks for your advice.

erikbor:
@jjr840430 sorry I didn't have time earlier to review this. I agree that a temporary file is not the best way because that currently can be done outside of Retrace. Moving that into Retrace itself does not improve things.

And I also agree with configuration setup flow that @ikolomi described.

ronaldtse:
@erikbor @jjr840430 are we ready to merge this?

jjr840430:
@ronaldtse I need to modify something here. I will let you know after doing.

jjr840430:
Currently `retrace` is parsing the configurations based on `file pointer`, so it needs to convert the most part of config parsing.
If we need to have new config by JSON, then this pull request may be ignored.

Issue #333, opened 2017-11-15
Development of a utility that can execute commands in parallel that are specified in a single command file


erikbor:
For the fuzzing efforts we'd like to have a utility that can execute commands in parallel/concurrently.

Usage:
```sh
$ spawn <timeout in seconds> <number of forks> <command file>
```

The content of the command file (e.g.: `/tmp/cmd.txt`) looks like:
```
command 1
command 2
command 3
command 4
command 5
command 6
command 7
command 8
command 9
command 10
command 11
command 12
```

We could execute this command file like this:
```sh
$ spawn 5 4 /tmp/cmd.txt
```

The utility would first execute `command 1`, `command 2`, `command 3`, `command 4` with a timeout of 5 seconds. Once a command has finished (whether it exited `0` or got terminated because of the timeout) the next command will be executed. At all times we should have a maximum of 4 commands running.

For now it is not important to display the output of the child processes. The output can be sent to `/dev/null`. It is also not important to check the return code of the child processes.

Btw, this code will eventually be integrated in the Retrace executable.

jjr840430:
@erikbor It was implemented by #336

ronaldtse:
Shipped: `tools/spawn` (`retrace-spawn`) launches N concurrent traced processes from one invocation.

PR #336, opened 2017-11-17
Added spawn utility

jjr840430:
Implemented the feature #333

### DIFF
--- a/tools/spawn/spawn.c
+++ b/tools/spawn/spawn.c
@@ -34,7 +34,6 @@
 #include <signal.h>
 #include <unistd.h>
 
-#define SPAWN_PROG_NAME                "spawn"
 #define MAX_TIMEOUT                    60          /* 60 seconds */
 
 /*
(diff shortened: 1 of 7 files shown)


ronaldtse:
@jjr840430 thanks for this. @erikbor 's intention is actually to have this integrated within retrace -- could the main code be put inside retrace, and the `spawn` script be just a simple C wrapper that includes this functionality? Thanks!

tools/spawn/spawn.c
```diff
@@ -0,0 +1,414 @@
...
+#include <stdlib.h>
+#include <pthread.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <fcntl.h>
+#include <signal.h>
+#include <unistd.h>
+
+#define SPAWN_PROG_NAME                "spawn"
```
drystone: Normally retrieved from `argv[0]`

tools/spawn/spawn.c
```diff
@@ -0,0 +1,414 @@
...
+		if (g_opt.fork_list[i].th < 0)
+			continue;
+
+		pthread_join(g_opt.fork_list[i].th, NULL);
+	}
+
+	/* free forking info list */
+	free(g_opt.fork_list);
+
+	/* destroy mutex */
```
drystone: Please avoid redundant comments ... this is mentioned in the coding guidelines.

tools/spawn/spawn.c
```diff
@@ -0,0 +1,414 @@
...
+		size_t len = 0, ret;
+
+		/* get command line */
+		ret = getline(&line, &len, cmd_fp);
+		if (ret < 0) {
+			if (line)
+				free(line);
+			break;
+		}
+		line[ret] = '\0';
```
drystone: `getline` does NULL termination for you

tools/spawn/spawn.c
```diff
@@ -0,0 +1,414 @@
...
+
+		/* chop endline */
+		if (strstr(line, "\r\n"))						/* windows format */
```
drystone: Not necessary unless this works on windows!

tools/spawn/spawn.c
```diff
@@ -0,0 +1,414 @@
...
+	if (g_opt.num_of_cmds == 0) {
+		fprintf(stderr, "Invalid command file '%s'\n", cmd_file);
+		exit(1);
+	}
+
+	/* get forks count */
+	g_opt.num_of_forks = strtoul(num_of_forks, NULL, 10);
+	if (g_opt.num_of_forks <= 0 || g_opt.num_of_forks > g_opt.num_of_cmds) {
+		fprintf(stderr, "The number of forks should be value in [1...%d]\n", g_opt.num_of_cmds);
+		exit(1);
```
drystone: I don't think it's an error if number of forks is less than commands. Either silently reduce `num_of_forks` or allow the spawn loop to catch this case.

tools/spawn/spawn.c
```diff
@@ -0,0 +1,414 @@
...
+			return -1;
+		}
+
+		/* set args */
+		args[tok_count] = tok;
+		tok_count++;
+
+		tok = strtok(NULL, " ");
+	}
+	args[tok_count] = NULL;
```
drystone: `args` can still be NULL here (I think a single space command line will trigger this bug)

tools/spawn/spawn.c
```diff
@@ -0,0 +1,414 @@
...
+
+	/* open /dev/null for redirecting output */
+	fd_null = open("/dev/null", O_WRONLY);
```
drystone: I think this will leak a file descriptor in the parent process. Best open `/dev/null/` in the child (after the fork)

jjr840430:
@drystone Thanks for your reviews. Those make sense to me.

By the way, I have some questions for your opinions about threading per fork and waitpid.
Did you look the requirements in #333?

The `command file` may contain the number of commands more than specified `number of forks`.
Let me know your opinions. Thanks.

drystone:
@jjr840430 I was thinking of something like (not complete or tested):

```
for (;;) {
  while (proc_reminaing > 0 && proc_count < max_proc_count) {
    fork_a_process();
    ++proc_count;
    --proc_remaining;
  }

  /* code here to kill long running processes */
  ...

  pid = waitpid(-1, &st, WNOHANG);
  if (pid > 0)
    if(--proc_count == 0 && proc_remaining == 0)
      break;

  sleep(1);
}
```
I've omitted code to terminate long running processes etc.

Even better would be to use something like `timer_create()` to deliver the signal to the long running process and replace `waitpid` with plain old `wait` (hence no need for `sleep`).

```
while (proc_remaining > 0 || proc_count > 0) {
  if (proc_remaining > 0 && proc_count < max_proc_count) {
    fork_a_process_with_sigkill_timer();
    ++proc_count;
    --proc_remaining;
  }
  else if (wait(NULL) > 0)
    --proc_count;
}
```

erikbor:
@jjr840430 Looks good so far!!!

Can you make a couple of modifications please:
1. Replace `prog_name` with `extern char *__progname;`
2. Move the code from `spawn.c` into a `.h` file and call those functions in a small `spawn.c` (this is to prepare `spawn` to be merged into Retrace itself)
3. Increase `MAX_TIMEOUT` to **3600** and change `if (g_opt.timeout < 0 || g_opt.timeout > 60) {` to `if (g_opt.timeout < 0 || g_opt.timeout > MAX_TIMEOUT) {` and turn `[0 ~ 60] seconds` into `[0 ~ %d] seconds", MAX_TIMEOUT`
4. Change `No left command to be executed at thread[#%d]` to `No command to be executed at thread[#%d]`
5. Change `The command '%s' has killed by timeout at thread[#%d]` to `The command '%s' was killed (timeout at thread[#%d])`
6. Change `Invalid command line '%s' at thread[#%d],` to `Invalid command '%s' at thread[#%d]`

And compilation on macOS shows this warning:
```
spawn.c:113:11: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
                if (ret < 0) {
                    ~~~ ^ ~
spawn.c:386:11: warning: incompatible integer to pointer conversion assigning to 'pthread_t' (aka 'struct _opaque_pthread_t *') from 'int' [-Wint-conversion]
                        fi->th = -1;
```

jjr840430:
@erikbor Can you check why `retrace-github-pr-macos` was failed?

erikbor:
@jjr840430 there's an issue with one of the macOS Jenkins slaves. No code issues 👍

jjr840430:
Thanks! :)

ronaldtse:
@erikbor since there are no further issues I'm going to merge this. Thanks @jjr840430 !

Issue #331, opened 2017-11-15
Replace the `retrace` shell script with an executable


erikbor:
Hi @jjr840430 

Can you replace the `retrace` shell script with an executable?

Can you implement a check whether the target binary has the SUID bit set and exit accordingly because we can't retrace a SUID binary.

This is part of the initial work to make Retrace suitable for automation.

Thanks!

ronaldtse:
Shipped with the v2 CLI (PR #497): an executable replaces the shell-script launcher entirely.

PR #337, opened 2017-11-20
Added wrapper tool

jjr840430:
Added simple utility to convert a shell script `retrace` to executable binary.(issue #331 )


jjr840430:
@drystone Here is the simple wrapper utility for `retrace`.
Do you think this implementation may be used for making an executable binary for `retrace`?

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+	fprintf(stderr, "Usage: %s -l <path of libretrace.so> <command line>\n",
+			prog_name);
+	exit(1);
+}
+
+/*
+ * fork process
+ */
+
+static void fork_process(int argc, char **argv)
```
drystone: `fork_process` will be more useful if it takes a single args parameter suitable for `execvp`. It turns out that `argv` in `main` is already a null terminated array ...

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+
+	/* set arguments */
+	args = (char **) malloc((argc - 1) * sizeof(char **));
+	if (!args) {
+		fprintf(stderr, "Out of memory!\n");
+		exit(1);
+	}
+
+	for (i = 0; i < argc - 3; i++)
+		args[i] = argv[i + 3];
```
drystone: ... so up to here is unnecessary

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+
+	/* open stdout */
+	dup2(STDOUT_FILENO, dup_stdout);
+	dup2(STDOUT_FILENO, dup_stderr);
```
drystone: This doesn't open anything useful. It simply duplicates fds 1 and 2 to `dup_stdout` and `dup_stderr` which are not initialised. Which will at best leak two fds and at worst redirect stderr to stdin or stdout (if `dup_stderr` happens to be 0 or 1)

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+
+	/* fork new process */
+	pid = fork();
+	if (pid == 0) {
+		dup2(STDOUT_FILENO, dup_stdout);
+		dup2(STDERR_FILENO, dup_stderr);
```
drystone: See above. `Fork`ed children inherit fds anyway so I'm not sure what you're trying to achieve here.

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+		fprintf(stderr, "Failed to fork new process(err:%d)\n", errno);
+		exit(1);
+	}
+
+	/* wait until the process has terminated */
+	do {
+		int w, status;
+
+		/* wait until process has terminated */
+		w = waitpid(pid, &status, -1);
```
drystone: -1 for `options` will enable all possible flags. I'm sure you mean `waitpid(pid, &status, 0)`.

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+		if (w < 0) {
+			fprintf(stderr, "waitpid() error for command\n");
+			break;
+		}
+
+		/* set exit status */
+		if (WIFEXITED(status))
+			break;
```
drystone: This test is pointless. Unless I've misunderstood your use of `waitpid`, the child has definitely exited.

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+	} while (1);
```
drystone: It will be more useful if `fork_process` returns the pid and `wait` happens elsewhere. If `fork_process` blocks, then you'll need another process/thread for additional tracees - remember the spawn utility?

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+int main(int argc, char *argv[])
+{
+	/* check arguments */
+	if (argc < 4)
+		usage(argv[0]);
+
+	if (strcmp(argv[1], "-l") != 0)
+		usage(argv[0]);
```
drystone: Rather use `getopt` here and use default locations for libretrace (probably /use/lib, /usr/local/lib) in case `-l <soname>` is not specified.

jjr840430:
@drystone It's just simple utility to show the ability to build `retrace` as executable binary.
So I didn't regard the details of full implementation.
If it's enough to improve building execution binary of `retrace`, then I will remove it from `tools` and build new binary in `src` directory.

Do you think it's the right way to implement executable binary of `retrace`?

tools/wrapper/retrace.c
jjr840430: @drystone Using `getopt` function is some complex because it needs to specify `command line`  to be executed.

drystone: This is a very common use case for `getopt` which is only interested in the option characters (and/or option names for `getopt_long`) and their respective arguments. The [man page](https://linux.die.net/man/3/getopt) has all the info.

tools/wrapper/retrace.c
```diff
@@ -0,0 +1,62 @@
...
+
+	fork_process(argv);
```
drystone: `argv + 3` to make `fork_process` more generic ... although will be `argv + optind` once you implement option parsing with `getopt`

tools/wrapper/retrace.c
jjr840430: Yes, I know about `getopt` function. But this utility has command line options to be executed.
It's not possible to be parsed by `getopt` properly.
If the syntax is specified in the following format, then using of 'getopt' is possible:

```
./retrace -l <library> -c <"command line">
```
In this case, the command line should be in `" "`. I think it's not preferred way.

drystone:
```
#include <unistd.h>
#include <stdio.h>

void
main(int argc, char *argv[])
{
        for (;;) {
                switch (getopt(argc, argv, "a::b:c")) {
                case 'a':
                        printf("getopt a: %s\n", optarg ? optarg : NULL);
                        break;
                case 'b':
                        printf("getopt b: %s\n", optarg);
                        break;
                case 'c':
                        printf("getopt c: (flag)\n");                    
                        break;
                default:
                        goto done_options;
                }
        }

done_options:
        while (optind < argc) {
                printf("argument: %s\n", argv[optind]);
                ++optind;
        }
}
```
```
$ make t
cc    -c -o t.o t.c
cc   t.o   -o t
$./t -aoptional -bmandatory -c execute me please
getopt a: optional
getopt b: mandatory
getopt c: (flag)
argument: execute
argument: me
argument: please
```

jjr840430: @drystone Assume the following example:
The first example:
```
./wrapper -l /bin/ls
```

The second example:
```
./wrapper /bin/ls -l
```
For both cases, `getopt` process `-l` option as `wrapper` utility.
But in the second case, `-l` option is for `/bin/ls`.

That's why I don't use `getopt` when parsing command line arguments.

drystone: There is the '+' mode for this. 
```
By default, getopt() permutes  the  con‐
tents of argv as it scans, so that even‐
tually all the  nonoptions  are  at  the
end.   Two  other  modes are also imple‐
mented.  If the first character of  opt‐
string  is  '+' or the environment vari‐
able POSIXLY_CORRECT is set, then option
processing  stops as soon as a nonoption
argument is encountered.
```

erikbor:
@jjr840430 great, can you implement the following please:

1. Check if `getuid` != 0 && `stat()` the target executable for setuid bit because if it is setuid then we can't LD_PRELOAD/DYLD
2. Check if the target executable is ELF or MACH
3. Replace `prog_name` with `extern char *__progname;` so `usage()` won't need an argument

ronaldtse:
@jjr840430 if you have time please feel free to complete this. Thanks!

jjr840430:
@ronaldtse seems ready to be merged.

ronaldtse:
@erikbor ready to merge this?

erikbor:
@jjr840430 can you please fix the following in `retrace.c`:

```C
"\t--lib <library path>     The path of libretrace.so library\n"
```
The `.so` message is now hardcoded but you did implement a `#ifndef __APPLE__`, so maybe it would be nice to reflect that back in the message. Thanks.

erikbor:
@jjr840430 I'm fine with merging! Looks good. Thanks a lot.

jjr840430:
@erikbor Thanks! :)

jjr840430:
@erikbor Fixed

ronaldtse:
Thanks @jjr840430 @erikbor !

PR #341, opened 2017-12-29
Fixed leak issues

jjr840430:


ronaldtse:
LGTM, let's merge this once @erikbor tries it out.

src/strinject.c
```diff
@@ -729,7 +729,7 @@ static size_t merge_iov_buffers(const struct iovec *iov, int iovcount, void **mb
 	for (i = 0; i < iovcount; i++) {
 		p = real_realloc(p, msize + iov[i].iov_len);
 		if (!p)
-			return -1;
+			return 0;
```
erikbor: Shouldn't we return NULL here instead?

PR #342, opened 2018-01-05
V1.0.2

jjr840430:


.travis.yml
```diff
@@ -1,5 +1,6 @@
 dist: trusty
-sudo: false
+sudo: true
+group: deprecated-2017Q4
```
ronaldtse: @jjr840430 could we use the latest container instead of the deprecated one?

jjr840430:
@ronaldtse I removed the line to use deprecated one.

PR #348, opened 2018-06-02
The following libralies can now be shimmed: ctype, locale, signal std…

ikolomi:
…io stdlib uio unistd. This commit is for Osx only


ronaldtse:
Thanks @ikolomi !

Issue #391, opened 2018-12-28
printf prints the wrong number for floats


riataman:
This program:

```
#include <stdio.h>
int main(void)
{
        printf ("%.0f %d\n", 1.0, 1);
}
```
will print "1 1" when run on its own, but when run under retrace v2 would print "0 1"

ronaldtse:
@riataman good find -- can we add a test for this?

riataman:
I need to revamp the tests, I don't think we are running any tests for v2 right now.

PR #392, opened 2018-12-28
Fixed issue when logging numbers

riataman:


ronaldtse:
@riataman Is #391 fixed with this?

riataman:
It's not, I fixed a minor issue when logging numbers here. I tried to fix #391 unsuccessfully. Will give it another try.

Issue #404, opened 2019-01-24
Running ps under retrace will crash


riataman:
Debian 9 x86.  Running `ps` under retrace v2 will run into this:

LD_PRELOAD=.libs/libretrace_v2.so ps
Signal 11 (SEGV) caught by ps (3.3.12).
ps:ps/display.c:66: please report this bug
Segmentation fault

riataman:
fixed by https://github.com/riboseinc/retrace/pull/405

PR #405, opened 2019-01-25
Fix crashes

riataman:
some null checks


riataman:
fixes https://github.com/riboseinc/retrace/issues/404

PR #413, opened 2019-02-04
Fix build in freebsd 12

riataman:


ronaldtse:
Thank you @riataman !

PR #427, opened 2022-01-18
MSYS configuration script

maxirmx:


maxirmx:
Hello @ribose-jeffreylau 
This PR includes the working windows configuration script. There were several issues, not onlt ACLOCAL path

HOWEVER, that was a minor  problem.  If you now look at build errors you will see that RTLD_NEXT is not defined.  dlsym with RTLD_NEXT option is the hooking method that dtrace uses. 
It is not a part of POSIX  and it is not available in MSYS. 
Frankly speaking, I have no idea (yet) how to fix it in rasonable time.

cc @ronaldtse

maxirmx:
Windows has system call that provides for similar implementation  --  SetWindowsHookEx
https://docs.microsoft.com/en-us/windows/win32/winmsg/using-hooks

There is mingw package that pretends to be a wrapper  
https://packages.msys2.org/package/mingw-w64-x86_64-dlfcn  but it is mingw and thus reuires windows oolchain that can not support other retrace features

ribose-jeffreylau:
Thanks @maxirmx !  I've distilled and pushed your fix into the `main` branch as af483aa8d01bf41a31e87a0229235517620d1171 .

PR #447, opened 2026-07-26
Enable v2 integration tests (issue #445 verified fixed)

### LINKED ISSUE #445: v2 segfaults on Linux when LD_PRELOADed into trivial test programs
## Summary

Discovered via Phase 8 test pyramid work (PR #444). When `libretrace_v2.so` is `LD_PRELOAD`ed into ANY of the 25+ test binaries (test/id, test/env, test/file, ...), the test binary segfaults (rc=139) on every Linux runner.

## How it was discovered

Previous `runtests.sh` used `if [ -f ../src/v2/.libs/libretrace_v2.so ]` — under the CMake build, that path never existed, so v2 tests silently skipped on every PR. PR #444 replaced the script with one that uses correct CMake paths and adds exit-code checking; the silent skip became a loud failure.

## Repro

```sh
cmake -B build -DRETRACE_BUILD_TESTS=ON && cmake --build build
LD_PRELOAD=$PWD/build/src/v2/libretrace_v2.so.2.0.1 build/test/id
# Segmentation fault (core dumped)
```

Affects every Linux runner in the matrix (ubuntu-22.04 x64+arm64, ubuntu-24.04 x64+arm64, ubuntu-26.04 x64+arm64).

macOS runners (Intel + arm64) also crash.

## Suspect

The v2 constructor (`src/core/main.c::retrace_main`) calls `retrace_real_impls_init` which on Linux calls:

```c
handle = retrace_real_impls.dlopen("libpthread.so.0", RTLD_NOW | RTLD_GLOBAL);
if (handle == NULL)
    return -2;
```

On modern glibc (>= 2.34) there is no separate `libpthread.so.0` — pthreads is part of libc. So `dlopen` returns NULL, init returns -2, `retrace_inited` stays 0.

Subsequent libc calls then go through `retrace_engine_wrapper` which calls `retrace_as_get_real_safe(func_name)` → on x86-64 this is `_dl_sym(RTLD_NEXT, ...)` (a GLIBC_PRIVATE symbol that may also be unavailable on modern glibc).

The fallback path (`if (!retrace_inited)`) calls `retrace_as_sched_real(arch_spec_ctx, real_impl)` — but if `real_impl` is NULL (because `_dl_sym` failed), the trampoline tries to jump to NULL → segfault.

## Tasks

- [ ] Fix `_dl_sym` resolution on glibc >= 2.34 (use `dlsym(RTLD_NEXT, ...)` as fallback, like the AArch64 backend does)
- [ ] Drop the `dlopen("libpthread.so.0")` check on glibc >= 2.34
- [ ] Verify all v2 tests pass on Linux x64+arm64 after fix

## Blocks

- v2.1.0 release
- Phase 8 test pyramid completion
- Phase 9 v1 removal (since v2 must work before v1 goes away)

ronaldtse:
## Summary

Three coupled changes that close #445 (v2 segfault on glibc >= 2.34):

1. **`src/backends/preload_elf/x86_64/arch_spec_bottom.c`** — declare `_dl_sym` as a weak symbol. On glibc >= 2.34 (where `_dl_sym` is no longer exported), it resolves to NULL; the code falls back to `dlsym(RTLD_NEXT, ...)`. Already in main via PR #444.

2. **`src/core/real_impls.c`** — stop failing init when `dlopen("libpthread.so.0")` returns NULL. On glibc >= 2.34, libpthread is integrated into libc; there's no separate libpthread.so.0 to dlopen. The previous code returned -2, leaving `retrace_inited=0`, which caused every libc call to fall through to `retrace_as_sched_real(..., NULL)` → segfault.

3. **`test/CMakeLists.txt` + `test/runtests.sh.in`** — remove the `RETRACE_SKIP_V2=1` skip and fix the `TEST_DIR` substitution (was using `$<TARGET_FILE_DIR:retrace_v2>` which gave `build/src/v2/`; tests live in `build/test/`). Now `configure_file(@ONLY)` substitutes `@TEST_DIR@` and `file(GENERATE)` substitutes `$<TARGET_FILE:retrace_v2>`.

Closes #445.

## Test plan

- [ ] Linux x64 + arm64: integration-runtests shows real PASS per test program (no longer SKIP, no longer segfault)
- [ ] macOS Intel + arm64: same
- [ ] Windows: smoke test still works (no v2 core on Windows)

### DIFF
--- a/src/core/real_impls.c
+++ b/src/core/real_impls.c
@@ -43,15 +43,21 @@ int retrace_real_impls_init(void)
 		return -1;
 
 #ifdef __linux__
-	/* load all libs that are used in safe mode,
-	 *  since the linker won't link them
+	/* On glibc < 2.34, pthreads lived in a separate libpthread.so.0 that
+	 * the dynamic linker wouldn't load unless explicitly requested.
+	 * Loading it made pthread_key_create etc. resolvable via RTLD_NEXT.
+	 *
+	 * On glibc >= 2.34, libpthread is integrated into libc -- there's no
+	 * separate libpthread.so.0 to dlopen. The dlopen call returns NULL,
+	 * and that's fine: the symbols are already in libc. Don't fail init
+	 * in that case.
 	 */
 	void *handle;
 
 	handle = retrace_real_impls.dlopen(
 			"libpthread.so.0", RTLD_NOW | RTLD_GLOBAL);
-	if (handle == NULL)
-		return -2;
+	/* handle may be NULL on glibc >= 2.34 -- not an error. */
+	(void)handle;
 #endif
 
 	retrace_real_impls.pthread_key_create =
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -58,9 +58,18 @@ endforeach()
 set(RETRACE_CONF_EXAMPLE "${CMAKE_SOURCE_DIR}/examples/getenv-fuzzing/retrace.conf.example")
 set(TEST_DIR "${CMAKE_BINARY_DIR}/test")
 
+# 1. configure_file(@ONLY) substitutes @TEST_DIR@ and @RETRACE_CONF_EXAMPLE@
+#    at configure time.
+# 2. file(GENERATE) then evaluates $<TARGET_FILE:retrace_v2> at build time
+#    (generator expressions aren't allowed in configure_file).
+configure_file(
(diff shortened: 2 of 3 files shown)


ronaldtse:
Closing this PR. The libpthread fix didn't resolve the segfault — it's deeper than init order. Issue #445 has the findings so far. Will require offline gdb debugging on Linux.

Issue #451, opened 2026-07-27
v2 printf/writev segfault on macOS arm64 under DYLD_INSERT_LIBRARIES


ronaldtse:
## Summary

Spun out of #445. PR #449 (asm-to-C `retrace_as_call_real`) fixed 18 of 24 v2 integration tests on macOS arm64. `printf` segfaults and `writev` returns rc=255.

## Repro

```sh
cmake -B build -G Ninja -DRETRACE_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release
cmake --build build
DYLD_INSERT_LIBRARIES=$PWD/build/src/v2/libretrace_v2.dylib ./build/test/printf
# Segmentation fault: 11
DYLD_INSERT_LIBRARIES=$PWD/build/src/v2/libretrace_v2.dylib ./build/test/writev
# rc=255
```

## Suspect

`printf` and `writev` are variadic. The plain-C dispatch in `src/core/as_call_real.c` uses fixed-arity function-pointer typedefs (`fn0` through `fn6`). Variadic functions on macOS arm64 may pass float/SIMD args differently than the dispatch assumes.

The asm version (now removed from per-backend `arch_spec_bottom.c`) had a `TODO: support float params` note, suggesting varargs/float handling was always incomplete.

## Workaround

Skipped in `test/runtests.sh.in` via the `should_skip` function (Darwin + `printf|writev`). Once fixed, remove the case.

## Acceptance criteria

- [ ] `DYLD_INSERT_LIBRARIES=libretrace_v2.dylib ./build/test/printf` returns 0
- [ ] `DYLD_INSERT_LIBRARIES=libretrace_v2.dylib ./build/test/writev` returns 0
- [ ] Remove `printf|writev` from `should_skip` in `test/runtests.sh.in`

PR #465, opened 2026-07-28
Fix aarch64 FP varargs segfault: use trampoline tail-call when no params modified

ronaldtse:
## Summary

Root-cause fix for the aarch64 FP varargs segfault class (issue #451, Alpine arm64 printf/scanf failures, and any case where `printf("%f", x)` was intercepted).

### Root cause

The trampoline has TWO ways to invoke the real libc implementation:

**PATH A (tail-call):** when `call_real_flag=1`, the trampoline restores `x0..x7` AND `v0..v7` (and `lr`, `sp`) from the saved frame, then tail-calls `real_impl`. The real function sees the original args in their original positions. `va_arg` works correctly for both integer and FP varargs.

**PATH B (C dispatch):** the `call_real` action calls `retrace_as_call_real` -> `retrace_as_call_real_dispatch`, which casts `real_impl` to a `(long, long, ...)` typedef and invokes it via plain C function pointer call. The compiler puts args in `x0..x7` only; `v0..v7` are NOT populated. `va_arg` for an FP arg reads garbage from `v0..v7` and typically segfaults.

The `call_real` action was always using PATH B — even for the common config `log_params + call_real` where no param modifications have been made.

### Fix

- Track whether any `modify_in_param_{str,int,arr}` action ran (new field `t_ctx->params_modified`, zeroed in `clear_context`).
- In `ia_call_real`, if `params_modified == 0`, set `call_real_flag=1` (PATH A) and skip the C dispatch. The trampoline tail-calls the real implementation with the original captured registers.
- If `params_modified == 1`, fall back to PATH B as before. This case can break FP varargs, but combining FP varargs with arg modification is rare; documented limitation.

Reverts the parser's "bail to 0 if format has FP specifiers" workaround in `printf_compat.h` — no longer needed.

## Test plan

- [ ] CI green on this branch
- [ ] Alpine x86_64: 21/21 tests pass
- [ ] Alpine aarch64: 21/21 tests pass (no skips needed)
- [ ] macOS arm64 still passes (and #451 may now be addressable in a follow-up)

## Follow-ups

- Remove the `printf|writev` skip from `runtests.sh.in` for Darwin arm64 (issue #451).
- The macOS backends have analogous trampoline+engine structure; verify PATH A is used there too.

### DIFF
--- a/src/core/printf_compat.h
+++ b/src/core/printf_compat.h
@@ -1,22 +1,22 @@
 /*
  * Compatibility shim for glibc's printf.h.
  *
- * glibc's printf.h provides:
- *   - the PA_* enum (PA_INT, PA_CHAR, ..., PA_LAST)
- *   - the PA_FLAG_* family (SHORT, LONG, LONG_LONG, LONG_DOUBLE, PTR, MASK)
+ * glibc's printf.h provides the PA_* enum, the PA_FLAG_* family, and
+ * parse_printf_format().
  *
- * These are used as integer markers by:
- *   - src/core/data_types.c    (lookup table from printf-arg-type to DataType)
- *   - src/backends/preload_elf/aarch64/arch_spec_bottom.c (varargs walk)
- *   - src/backends/preload_macho/* (custom printf domain registration;
- *     register_printf_domain_function is only available on glibc.)
+ * Used as integer markers by data_types.c (lookup table from
+ * printf-arg-type to DataType), preload_elf arch_spec_bottom.c
+ * (varargs walk: parse fmt, read the right number of args from the
+ * captured frame, pass them through dispatch to the real libc impl),
+ * and preload_macho backends (custom printf domain registration;
+ * register_printf_domain_function is only available on glibc/Darwin).
  *
  * musl (OHOS, Alpine) does not ship printf.h and does not provide
  * register_printf_*. The constants are still needed as integer markers,
  * so we define them locally when the system header is absent.
  *
- * Values mirror glibc exactly so any argtype produced by parse_printf_format
- * on a glibc host still indexes correctly.
+ * Values mirror glibc exactly so any argtype produced by
+ * parse_printf_format on a glibc host still indexes correctly.
  */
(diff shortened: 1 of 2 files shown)


ronaldtse:
Reverted — the PATH A approach breaks macOS arm64 (setlocale SIGTRAP).

## Investigation findings

**Root cause of FP varargs failures:**
The trampoline captures x0..x7 AND v0..v7 in the frame. But the C-dispatch path (retrace_as_call_real_dispatch) casts to (long, long, ...) and calls via function pointer -- the compiler puts all args in x0..x7 only. FP varargs that the caller put in v0..v7 are invisible to the dispatch.

**Attempted fix (this PR):**
Use the trampoline's tail-call path (PATH A) when no params are modified. PATH A restores both x0..x7 AND v0..v7 before branching to real_impl.

**Result:**
- Alpine x86_64 + aarch64: PASS (all 21 tests)
- macOS arm64 (14, 15, 26): FAIL -- setlocale SIGTRAP (rc=133)

The macOS failure is likely due to Apple Silicon arm64e pointer authentication or dyld interpose semantics interacting with the trampoline's `br x16` instruction. Debugging requires local macOS hardware.

## Current state (after revert)

PR #464 (already merged) handles the practical cases:
- printf("%d %s", ...) and scanf("%s", ...) work correctly via parse_printf_format
- printf("%f", ...) uses the FP bail (returns 0, falls back to named-args-only)
- All Alpine tests pass (21/21 on both x86_64 and aarch64)
- macOS arm64 still skips printf/writev (issue #451)

## What the proper fix requires

1. Make wrapper_frame_get_arg type-aware: for PA_DOUBLE args, read from frame->real_v0..v7 instead of frame->real_x0..x7
2. Make retrace_as_call_real_dispatch FP-aware: use a union type or assembly dispatch that places FP args in v0..v7 and int args in x0..x7
3. Investigate macOS arm64 PATH A failure: test with PAC disabled or use braaz instead of br

These are deeper trampoline/engine changes that need careful per-platform debugging.

PR #466, opened 2026-07-28
Fix FP varargs: trampoline blr + reentrance-aware cleanup (#451)

ronaldtse:
## Summary

Full root-cause fix for the aarch64 FP varargs segfault class (issue #451, `printf("%f", x)` crashes, Alpine arm64 printf/scanf failures).

### Root cause

`retrace_as_call_real_dispatch` casts the real function pointer to `(long, long, ...)` and calls via plain C function pointer. On aarch64 AAPCS64 the compiler places all args in x0..x7 only; FP varargs that the caller put in v0..v7 are invisible. The callee's `va_arg` reads garbage and segfaults.

### Fix

When no params have been modified (the common case: `log_params + call_real`), defer to the trampoline's "call real + cleanup" path. The trampoline:

1. Restores the original x0..x7 AND v0..v7 from the saved frame
2. Calls real_impl via `blr` (not `br` — we need control back)
3. After real_impl returns, calls `retrace_path_a_cleanup` to decrement the reentrance depth
4. Returns real_impl's return value to the original caller

### Reentrance handling (the key insight from lldb debugging)

The previous attempt (PR #465) used `br` (tail-call) which crashed on macOS with `SIGTRAP` inside `_os_unfair_lock_recursive_abort`. Root cause: `setlocale` internally calls `strcpy`, which is also intercepted. With `br` (tail-call), `clear_context` ran before `real_setlocale` executed, clearing the reentrance guard. The nested `strcpy` call saw no active guard, tried to log, and deadlocked on macOS's non-recursive `os_unfair_lock` inside `localeconv_l`.

The new approach uses `blr` + a depth counter:
- `clear_context` preserves `real_impl` and sets `path_a_depth=1` when PATH A is chosen
- The engine's early-return (reentrance) path increments `path_a_depth`
- `retrace_path_a_cleanup` decrements depth on each nested return
- The guard is only cleared when depth hits 0

### Verified locally on Apple M1 Max (arm64)

- All 20 tests pass (writev skipped for /dev/urandom issue)
- `printf("%f %lf %e", 3.14f, 2.71828, 1.23e-4)` → correct FP output
- `printf("%d %f %s %f", 42, 3.14, "hello", 9.99)` → mixed int/FP/string works
- malloc test (1000 iterations + FP printf at end) passes
- No reentrance deadlocks (setlocale → strcpy chain works)

### Files changed

- `engine.h`: add `params_modified` and `path_a_depth` to `ThreadContext`
- `engine.c`: `clear_context` preserves `real_impl` for PATH A; reentrance early-return increments depth; new `retrace_path_a_cleanup` called by the trampoline
- `actions/basic.c`: `call_real` defers to PATH A when `params_modified == 0`; `modify_in_param_*` sets the flag
- `arch_spec_top.S` (ELF + Mach-O): PATH A uses `blr` + cleanup instead of `br` (tail-call)
- `printf_compat.h`: remove FP bail (no longer needed)
- `runtests.sh.in`: remove `printf` skip on Darwin arm64 (issue #451 fixed); keep `writev` skip (/dev/urandom test issue)

## Test plan

- [ ] CI green on all platforms
- [ ] macOS arm64: printf test passes (was skipped, issue #451)
- [ ] Alpine x86_64 + aarch64: all tests pass
- [ ] Linux: no regressions

### DIFF
--- a/src/backends/preload_elf/aarch64/arch_spec_top.S
+++ b/src/backends/preload_elf/aarch64/arch_spec_top.S
@@ -141,10 +141,13 @@
 	ldr     x9, [sp, #OFFS_CALL_REAL_FLAG]
 	cbz     x9, 1f
 
-	/* call_real_flag != 0: restore arg regs and tail-call real_impl.
-	 * The real function's `ret` returns directly to the original caller
-	 * (x30 was saved in the frame and we restore it here). */
-	ldr     x16, [sp, #OFFS_REAL_IMPL]   /* branch target */
+	/* PATH A: call real_impl with the original args, then clean up
+	 * the reentrance guard. We use blr (not br) so we get control
+	 * back after real_impl returns; then we call
+	 * retrace_path_a_cleanup to decrement the depth and return
+	 * real_impl's value to the original caller.
+	 */
+	ldr     x16, [sp, #OFFS_REAL_IMPL]
 
 	ldp     x0, x1, [sp, #OFFS_REAL_X0]
 	ldp     x2, x3, [sp, #OFFS_REAL_X2]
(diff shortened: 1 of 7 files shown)


ronaldtse:
Reverted -- the blr + reentrance-depth approach causes infinite recursion on all arm64 platforms due to the reentrance guard not being visible during nested engine calls (retrace_func_get -> log_dbg -> vsnprintf -> memmove chain). Needs deeper investigation of the TLS guard mechanism. The parser improvement from PR #464 (already merged) handles the practical non-FP cases.

PR #493, opened 2026-07-29
fix: dlopen reentrance guard for Linux (#450)

### LINKED ISSUE #450: v2 segfaults when test/dlopen runs under LD_PRELOAD on Linux
## Summary

Spun out of #445. With the asm-to-C `retrace_as_call_real` fix from PR #449 in place, 20 of 24 v2 integration tests pass on Linux. `test/dlopen` is one of the 4 that still segfault.

## Repro

```sh
cmake -B build -G Ninja -DRETRACE_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release
cmake --build build
LD_PRELOAD=$PWD/build/src/v2/libretrace_v2.so.2.0.1 ./build/test/dlopen
# [retrace log output, ending mid-stream with]
# "Running action call_real, for malloc:(nil), tpid ..."
# Segmentation fault (core dumped)
```

## What's happening

`test/dlopen.c` calls `dlopen(NULL, RTLD_LAZY)`, then `dlsym(handle, "test")`, then `dlclose(handle)`. The dl* functions are NOT intercepted (PR #449 removed them). The segfault appears during the call_real action for `malloc` — which IS intercepted — when called from inside libc's dlopen implementation.

Likely root cause: `retrace_as_call_real_dispatch` (src/core/as_call_real.c) does `retrace_real_impls.malloc(sizeof(long) * params_cnt)` before dispatching. When the call comes from inside libc's dlopen path, the malloc trampoline recurses through the engine, and somewhere in that recursion the dispatch ends up calling a NULL or invalid pointer.

## Workaround

Skipped in `test/runtests.sh.in` via the `should_skip` function. Once fixed, remove the `dlopen) return 0 ;;` line.

## Acceptance criteria

- [ ] `LD_PRELOAD=libretrace_v2.so ./build/test/dlopen` returns 0
- [ ] Remove `dlopen` from `should_skip` in `test/runtests.sh.in`

ronaldtse:
## fix: dlopen reentrance guard for Linux (#450)

Closes #450.

### What this adds

New file `src/core/dlopen_guard.c` that exports `dlopen`/`dlclose`/`dlsym`/`dlerror` as global symbols in libretrace.so on Linux. Each shim sets a thread-local `__thread int g_retrace_in_dlopen` flag before calling real libc, and clears it after.

The engine checks `retrace_dlopen_guard_active()` at entry. If active (thread is inside dlopen's internal path), the engine skips all action processing and schedules real. The asm trampoline tail-calls real. No log entries for calls that happen inside dlopen.

This prevents the state inconsistency crash where libc's dlopen internals call `malloc`/`calloc`/`free` through our wrappers, triggering the engine, which somehow corrupts libc's internal state.

### Platform scope

- **Linux (ELF)**: shims compiled and exported via LD_PRELOAD symbol interposition. Full guard active.
- **macOS**: `dlopen_guard.c` NOT compiled (CMake `RETRACE_PLATFORM_LINUX` guard). The dlopen crash was Linux-specific. macOS tests unaffected.

### Also

- Adds `dlclose` to `retrace_real_impls` struct (was missing).
- Removes `dlopen` from `should_skip` in `test/runtests.sh.in`.

### Verification

- ctest green on macOS arm64 (unaffected).
- Linux x86_64 + arm64 verification pending CI — this is the critical test.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -11,6 +11,14 @@ set(_core_sources
 	engine.c        funcs.c         actions.c      data_types.c
 	real_impls.c    logger.c        main.c         as_call_real.c)
 
+# dlopen_guard.c is Linux-only: it exports dlopen/dlclose/dlsym/dlerror
+# as global symbols for LD_PRELOAD interposition. On macOS these symbols
+# cause dyld init issues. On macOS the guard functions are still declared
+# (in engine.h) but always return 0 (no-op).
+if(RETRACE_PLATFORM_LINUX)
+	list(APPEND _core_sources dlopen_guard.c)
+endif()
+
 set(_core_prototype_sources
 	prototypes/locale.c prototypes/ctype.c  prototypes/signal.c
 	prototypes/stdio.c  prototypes/stdlib.c prototypes/uio.c
--- a/src/core/dlopen_guard.c
+++ b/src/core/dlopen_guard.c
@@ -1,0 +1,152 @@
+/*
+ * Copyright (c) 2017, [Ribose Inc](https://www.ribose.com).
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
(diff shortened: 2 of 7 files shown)


ronaldtse:
Superseded by PR using asm trampoline approach (much simpler).

PR #505, opened 2026-07-30
ci: smoke test verifies interception actually took effect (#504)

### LINKED ISSUE #504: Smoke test doesn't verify interception on macOS (SIP strips DYLD_INSERT_LIBRARIES)
## Context

The smoke test in `.github/workflows/build.yml` (lines 147-162) runs:

```sh
DYLD_INSERT_LIBRARIES="$(pwd)/$lib" /usr/bin/id
LD_PRELOAD="$(pwd)/$lib" /usr/bin/id
```

On macOS, `/usr/bin/id` is SIP-protected. The kernel strips `DYLD_INSERT_LIBRARIES` for SIP-protected binaries, so the retrace library never loads. The smoke test "passes" because `/usr/bin/id` runs successfully and produces output, but no interception actually happens.

## What this means

CI for macOS Intel (PRs #498, #502) shows the smoke test passing, but this proves only that the library loads — not that interception works. The integration tests (`test/runtests.sh`) do run programs under v2 and check exit codes, but those programs also don't assert that interception happened (just that they didn't crash).

## Fix

The smoke test should:

1. Build a small test binary that uses a non-SIP path.
2. Run it under v2 with a config that flips observable behavior (e.g., `modify_return_value_int` on `getuid`).
3. Assert that the modified behavior took effect.

The `examples/id-redirection/` example provides exactly this pattern. Wire it into the smoke step.

Alternative: install the binary to `/tmp` (outside SIP-protected dirs) before running.

Refs #479, #478, #452.

ronaldtse:
## Summary

- Smoke test in `build.yml` now builds a tiny `getuid()` binary, redirects via JSON config, and asserts the output changes — closing the SIP loophole where macOS `/usr/bin/id` silently skips `DYLD_INSERT_LIBRARIES`.
- Verified locally on Apple Silicon: output is `uid=0` as expected.

## Test plan

- [x] Local run on macOS arm64 produces `interception verified`.
- [ ] GHA confirms step passes on every POSIX platform including Intel macOS.

Closes #504.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -145,25 +145,61 @@ jobs:
         run: ctest --test-dir build --output-on-failure
 
       # ----- Smoke tests (POSIX only — no LD_PRELOAD on Windows) -----
-      # Best-effort: skipped if the v2 library wasn't produced.
-      # The CMake target is `retrace_v2` for historical reasons but the
-      # output library is `libretrace.{so,dylib}` (Phase 9 / ADR-0011).
-      # Uses /usr/bin/id (works on both Linux and macOS; /bin/id does
-      # not exist on newer macOS).
+      # Two-step verification:
+      #   1. Load: library loads cleanly under LD_PRELOAD/DYLD_INSERT_LIBRARIES.
+      #   2. Intercept: a config that flips observable behavior actually
+      #      takes effect. This catches regressions where the library
+      #      loads but doesn't intercept (issue #504 — SIP hides this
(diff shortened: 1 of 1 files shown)


ronaldtse:
Closing — the smoke test correctly identified that interception still doesn't work on Intel macOS despite PR #498. The output was `uid=501` instead of `uid=0`, meaning `getuid()` was not intercepted. Filing a separate issue to track the underlying interception bug; will reland this smoke test once Intel actually passes it.

PR #513, opened 2026-07-30
real_impls_init: table-driven DRY refactor (257 → 143 LOC)

ronaldtse:
## Summary

Replaces the 35+ duplicated symbol lookups in `retrace_real_impls_init` with a single declarative table. DRY win — adding a new libc dependency goes from 2 lines of boilerplate to 1 line of table entry.

Behavior preserved exactly:
- `dlopen` resolves first (Linux libpthread load depends on it)
- `dlclose` stays non-required (NULL is fine on some platforms)
- All other 33 symbols required (NULL → init fails)
- Return codes match the original numbering

## Test plan

- [x] Fresh `cmake -B build -DRETRACE_BUILD_TESTS=ON && cmake --build build` clean, no new warnings.
- [x] `ctest --test-dir build` 4/4 pass (integration + 3 unit suites).
- [x] `RETRACE_JSON_CONFIG=getuid-redirect.json` → `uid=0` (interception end-to-end).
- [ ] GHA matrix green.

### DIFF
--- a/src/core/real_impls.c
+++ b/src/core/real_impls.c
@@ -32,182 +32,113 @@
 
 #include "real_impls.h"
 #include "arch_spec.h"
+#include "logger.h"
 
 struct RetraceRealImpls retrace_real_impls;
+
+/*
+ * Each entry maps a libc symbol name to its slot in retrace_real_impls.
+ * `required` mirrors the original semantics: if the lookup returns NULL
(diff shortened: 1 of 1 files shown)


ronaldtse:
Closing — the table-driven refactor causes a load-time segfault on OHOS (qemu/musl) that I can't reproduce locally. The DRY win isn't worth the regression. Reverting; the original direct-assignment code stays.

PR #631, opened 2026-08-14
release: v2.3.7 — parson NULL-input fix + budget regression tests

ronaldtse:
## Summary

Bumps retrace from v2.3.6 to **v2.3.7** (PATCH per ADR-0006). The new test caught a real NULL-input segfault bug in the first 30 seconds.

### Fixed — real bug found by writing the test

- `src/config/json/parson.c`: `json_parse_string_with_comments(NULL)` called `strlen(NULL)` and segfaulted. Any caller that passed NULL by accident (e.g. a config-file reader that returned NULL on EOF) would crash the entire process. Now returns NULL gracefully before `strlen`. The NULL-input test in the new `test_parson_budget.c` caught this within seconds of first running.

### Code quality

- `src/config/json/parson.c`: removed duplicate `static size_t parson_alloc_total;` and `parson_alloc_budget;` declarations. Both pairs initialized to 0 (C tentative definition semantics), so behavior was unchanged, but the duplication was a clear bad-merge artifact from the v2.3.0 OOM fix.

### Added — 10 new regression tests

- `test/unit/test_parson_budget.c` — guards the v2.3.0 allocation-budget mechanism against regressions:
  - Realistic configs parse (regression guard against over-tight budget).
  - Configs with comments parse (the path that originally OOM'd).
  - Minimal `{}` and `[]` parse.
  - Repeated parses don't accumulate (budget resets per call).
  - NULL input returns NULL gracefully (the bug above).
  - Empty string, whitespace-only, malformed JSON all return NULL gracefully.
  - Large realistic config (8 KB, 50 functions) stays within budget.

### Version + packaging

- `version.h`, `retrace_cli.c` banner: 2.3.6 → 2.3.7
- `packaging/{nix,debian,fedora}` bumped
- `CHANGELOG.md` entry

## Why this matters

The NULL-input bug was latent — it would only fire if some caller accidentally passed NULL (e.g. a config-file reader returning NULL on EOF). Without writing the budget regression test, the bug would have stayed latent until production. Writing the test caught it within seconds.

The budget mechanism itself now has a regression guard: any future change that tightens the budget too aggressively (breaking real configs) or fails to reset it between calls (breaking repeated parses) will fail this test.

## Test plan

- [x] `cmake --build build`
- [x] `ctest --test-dir build` — 50/50 pass (was 49)
- [x] `./build/test/unit/test_parson_budget` — 10/10 pass
- [x] `./build/test/property/property_parson` — all properties pass (no regression from the NULL fix)
- [x] `./build/src/cli/retrace --help` banner reads `v2.3.7`
- [ ] CI green on all matrix legs
- [ ] After merge: tag v2.3.7; release.yml produces binary artifacts

## Tagging plan

After PR rebase-merges, tag `v2.3.7` at the merge commit. release.yml will produce per-platform binary artifacts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 (see `docs/adr/0006-semantic-versioning.md`).
 
+## [2.3.7] - 2026-08-14
+
+### Fixed
+- `src/config/json/parson.c`: `json_parse_string_with_comments(NULL)`
+  no longer segfaults. The public API function called `strlen(NULL)`
+  before any NULL check, which would crash any caller that passed
+  NULL by accident. Now returns NULL gracefully. Bug found while
+  writing the budget regression test.
+
+### Changed
+- `src/config/json/parson.c`: removed duplicate `static size_t
+  parson_alloc_total;` and `parson_alloc_budget;` declarations.
+  Both pairs initialized to 0 (C tentative definition semantics),
+  so behavior was unchanged, but the duplication was a clear bad-
+  merge artifact.
+
+### Added
+- `test/unit/test_parson_budget.c` — 10 unit tests guarding the
+  v2.3.0 allocation-budget mechanism against regressions: realistic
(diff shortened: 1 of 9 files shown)


ronaldtse:
Superseded: checkpatch checks each commit in the PR range individually, so the whitespace fix needed to be part of the original commit. Reopening as a single clean commit.

PR #636, opened 2026-08-14
v2.4.0: native process attach via ptrace

ronaldtse:
## Summary

Bumps retrace from v2.3.8 to **v2.4.0** (MINOR per ADR-0006: new capability, backwards-compatible). Native process attach — `retrace attach <pid>` — reaching the targets `LD_PRELOAD` structurally cannot: any running PID, no restart, no control of the launch.

### `retrace attach <pid>` (Linux)

Attaches via ptrace and traces syscalls until the target exits. Same JSON output as `retrace run`; feeds the same downstream tools (audit, diff, replay). Clean failure elsewhere: NOENT when the ptrace backend isn't compiled in; PERMISSION with ptrace_scope guidance.

### `retrace backends`

Lists the interposition backends compiled into the library (`preload-elf`, `preload-macho`, `preload-msvc`, `ptrace`, ...).

### New public API (`include/retrace/retrace.h`)

- `retrace_attach_process(pid)` — explicit ptrace lookup, bypassing the static-binary probe. Attach semantics differ from spawn: for a process you cannot exec, ptrace is the sole native mechanism regardless of how the binary was linked.
- `retrace_list_backends(&names, &count)`.

### Implementation notes

- `registry.c` owns both new entry points (it owns backend lookup).
- `trace_loop.c`: removed a NULL-engine early return that defended an unsatisfiable contract — `struct retrace_engine` is never instantiated; syscall dispatch goes through the process-global `retrace_engine_wrapper`. The handle survives in the signature for the backend API contract.
- CLI dlopens the library it already locates (`find_library`) and dlsyms the new symbols; config env set before dlopen so the constructor picks it up — same contract as `run`.

### Tests (`test/unit/test_attach.c`, 3 cases)

- Backend enumeration returns ≥ 1 registered, names valid.
- Invalid pid (0, -1) rejected.
- **Linux**: fork child + PTRACE_ATTACH + alarm-kill round trip verifying the trace loop runs to completion; skips gracefully on ptrace-permission denial. **Non-Linux**: the API fails cleanly (NOENT/UNSUPPORTED), never OK. Verified 3/3 on macOS locally; the Linux legs of the CI matrix exercise the real attach path.

### Docs

- `cli.md`: sections for both subcommands (with ptrace-permission troubleshooting).
- `README.adoc`: What's-new v2.4.0 + ecosystem table row.
- Cookbook 29 (Frida): notes native attach next to the Frida path — Frida remains the answer for non-child processes under locked-down ptrace, iOS/macOS.

## Test plan

- [x] `cmake --build build`
- [x] `ctest --test-dir build` — 51/51 pass (was 50)
- [x] `./build/test/unit/test_attach` — 3/3 pass on macOS (Linux attach leg runs in CI)
- [x] `retrace backends` — clean output, lists `preload-macho` on macOS
- [x] `retrace attach <pid>` — clean "ptrace backend not available" on macOS, exit 1
- [x] `./build/src/cli/retrace --help` banner reads `v2.4.0`
- [ ] CI green on all matrix legs (Linux legs exercise the real attach)
- [ ] After merge: tag v2.4.0; release.yml produces binary artifacts

## Tagging plan

After PR rebase-merges, tag `v2.4.0` at the merge commit. release.yml will produce per-platform binary artifacts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 (see `docs/adr/0006-semantic-versioning.md`).
 
+## [2.4.0] - 2026-08-14
+
+MINOR bump per ADR-0006: new capability, backwards-compatible API.
+
+### Added
+- **Native process attach — `retrace attach <pid>`**. Attach to an
+  already-running process via ptrace and trace its syscalls until
+  it exits. No `LD_PRELOAD`, no restart, no control of the launch
+  required — reaches the targets the preload backends structurally
+  cannot (any running PID; static binaries after they started).
+  Output is the same JSON format as `retrace run` and feeds the
+  same downstream tools (audit, diff, replay). Linux only; reports
+  a clean error elsewhere.
+- **`retrace backends`** — lists the interposition backends
+  compiled into the library (preload-elf, preload-macho,
+  preload-msvc, ptrace, ...).
+- **New public API** (`include/retrace/retrace.h`):
+  `retrace_attach_process(pid)` — explicit ptrace lookup (bypasses
+  the static-binary probe: attach semantics differ from spawn; for
+  a process you cannot exec, ptrace is the sole native mechanism
+  regardless of how the binary was linked) — and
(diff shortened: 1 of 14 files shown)


ronaldtse:
Superseded: the 3 remaining warnings were the local fn-ptr declarations (casts were already named). Fixed in the reopened PR.

PR #645, opened 2026-08-17
v2.5.0: public API matches the implementation (ADR-0014)

ronaldtse:
## Summary

Bumps retrace from v2.4.6 to **v2.5.0** (MINOR per ADR-0006: the public header surface changed). The public API now matches the implementation — ADR-0014.

## The problem

An `nm -g` audit of the built library showed that of ~28 `RETRACE_API` functions declared in `<retrace/retrace.h>`, **only the two added in v2.4.0 actually existed** — even `retrace_version()` had no definition. The header claimed "ABI-stable from v2.0.0" while consumers who wrote against it compiled fine and failed at link time: the worst failure mode for a public interface, invisible until it's expensive.

### Removed — the phantom declarations

The never-implemented scaffold (engine lifecycle, script builder, action params, config parsing, introspection, error reporting) is removed from the header. **No working program can regress**: the symbols never existed in any linkable build. The shipped surface is exactly the exported symbols, verified by `nm`.

### Added

- `retrace_version()` + `retrace_version_info()` — implemented for real in a new `src/core/public_api.c` (previously declared, never defined).
- `test/unit/test_public_api.c` — the **surface guard**: dlsyms every function the header declares from the running image and fails the build if any is missing. A phantom declaration can never ship again. Also pins the version contracts and an attach/list-backends behavior smoke.
- `docs/adr/0014-public-api-matches-implementation.md` — the decision, alternatives considered (implement-all-now rejected: needs the engine-object redesign; keep-with-comment rejected: the failure mode remains), and the implementation-first re-introduction path. ADR index updated (0013 was also missing from it).

### Version + packaging

- `version.h`, `retrace_cli.c` banner: 2.4.6 → 2.5.0; packaging bumped; CHANGELOG entry.

## Test plan

- [x] `cmake --build build`
- [x] `ctest --test-dir build` — 57/57 pass (was 56)
- [x] `./build/test/unit/test_public_api` — 5/5 pass
- [x] `nm -g` on the built library: every declared symbol exports (8 `retrace_*` T-symbols)
- [x] No consumers of the removed names anywhere in src/test/tools/include
- [x] Single commit; verified committed content via `git show HEAD:`
- [ ] CI green on all matrix legs
- [ ] After merge: tag v2.5.0; release.yml produces binary artifacts

## Tagging plan

After PR rebase-merges, tag `v2.5.0` at the merge commit. release.yml will produce per-platform binary artifacts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,34 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 (see `docs/adr/0006-semantic-versioning.md`).
 
+## [2.5.0] - 2026-08-17
+
+MINOR bump per ADR-0006: the public header surface changed.
+
+### Changed
+- **The public API now matches the implementation** (ADR-0014).
+  An audit (`nm -g` on the built library) showed that of ~28
+  functions declared in `<retrace/retrace.h>`, only the two added
+  in v2.4.0 actually existed — even `retrace_version()` had no
+  definition. Consumers compiled against the header and failed at
+  link time. The never-implemented declarations (engine
+  lifecycle, script builder, action params, config parsing,
+  introspection, error reporting) are removed; the re-
+  introduction path is documented in the ADR and gated on real
+  implementations. No working program can regress: the removed
+  symbols never existed in any linkable build.
+
+### Added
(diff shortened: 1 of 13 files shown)


ronaldtse:
Superseded: the surface-guard test needs ENABLE_EXPORTS for dlsym(self) on Linux (macOS exports exe symbols by default; glibc needs --export-dynamic). Fixed in the reopened PR.

PR #651, opened 2026-08-18
v2.6.0: public registry introspection API + working list-* CLI

ronaldtse:
## Summary

MINOR per ADR-0006: public surface grows — the first slice of ADR-0014's implementation-first re-introduction.

### New public API

- **`retrace_list_functions()`** — every interceptable libc function (the prototype registry).
- **`retrace_list_actions()`** — every built-in action.
- Ownership contract matches `retrace_list_backends`. Declared with the definitions per ADR-0014; the dlsym surface guard now enforces both (8 symbols).

Implementation walks the `__retrace_funcs`/`__retrace_acts` linker-section arrays — no init required. The section macro's block-scope externs get hoisted by clang, so each listing lives in its own TU (mirroring funcs.c/actions.c) with a shared inline builder in `public_api_internal.h`.

### Fixed (user-visible): the list-* CLI works now

Both subcommands were stubs since the CLI landed: `list-functions` printed TODO text; `list-actions` printed a hardcoded list of 9 that had drifted from the real 17. Both now dlopen the installed library and print the live registry — **322 functions, 17 actions** verified.

### Tests

Surface guard extended + two contract tests (counts, non-empty names, well-known members). 7/7 in `test_public_api`; 59/59 overall.

## Test plan

- [x] `ctest --test-dir build` — 59/59
- [x] `retrace list-actions` → 17 entries; `retrace list-functions` → 322 entries (live registry)
- [x] Surface guard: 8/8 symbols resolve
- [x] Single commit; verified via `git show HEAD:` (stub text gone)
- [x] No AI attribution
- [ ] CI green
- [ ] After merge: tag v2.6.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,35 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 (see `docs/adr/0006-semantic-versioning.md`).
 
+## [2.6.0] - 2026-08-18
+
+MINOR bump per ADR-0006: public surface grows (ADR-0014's
+implementation-first path).
+
+### Added
+- **`retrace_list_functions()` / `retrace_list_actions()`** —
+  public registry introspection: enumerate every interceptable
+  libc function (the prototype registry) and every built-in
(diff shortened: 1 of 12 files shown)


ronaldtse:
Superseded: gcc 14+ rejects the const-mismatched cache pointer clang tolerated; helper signatures fixed in the reopened PR.

PR #688, opened 2026-08-24
ci: TODO 28 discriminator -- NtCreateFile alone, no logging

ronaldtse:
One-line CI discriminator for the `NtCreateFile` correctness defect (TODO.trace-profile/28): the v2.31.2 prologue dump showed the original stub is an 8-byte KVAS `test` that the disassembler accepted whole — patch mechanics sound — so this round runs `NtCreateFile` alone with a `call_real`-only script (no param serialization). Pass ⇒ serialization is the killer; fail ⇒ the trampoline return path. No product change.


ronaldtse:
Superseded by #689, which stacks this commit plus the round-3 evidence (failure-detail verdict + param-shape control). Merging #689 alone applies both.

PR #689, opened 2026-08-24
ci: TODO 28 round 3 -- failure-detail verdict + param-shape control

ronaldtse:
TODO 28 evidence round 3: the no-logging run also fails (rc 3), clearing param serialization — the minimal call_real path itself is broken for `NtCreateFile`. This round adds two evidence axes: the smoke target's verdict now prints `errno`/`GetLastError` on the failed `fopen`, and a param-shape control runs `NtOpenFile` alone no-log (6 params — it passed even *with* logging in the bisect) against `NtCreateFile` (11 params including the OUT `IO_STATUS_BLOCK`). No product behavior change beyond the target's evidence print.


ronaldtse:
Superseded: the round-4 branch (fix/ntcreatefile-round4) carries rounds 2-4 on a clean base (main, without the unrelated v2.32.0 work) — PR follows.

